PR #6, opened 2014-02-20
Add support for hstore

bullno1:
Currently typespec is:

    hstore() :: [{key(), value()}] | {[key(), value()}.
    key() :: atom() | binary() | string() | number().
    value() :: null | atom() | binary() | string() | number().

"jiffy-like" format is required for array support since there's no trivial way to distinguish between an array of hstore and a hstore.
Return type for hstore is always: [{binary(), binary() | null}] for convenience. Maps in 17.0 will solve this problem though.

Opened a pull request because I need some discussion about this.


davidw:
We have hstore support now.  Thanks!

PR #18, opened 2014-04-30
Convert from rebar to erlang.mk

bipthelin:
This project will only compile if being included from another project using rebar.

This patch converts the current Makefile/rebar.config into it's erlang.mk equivalent.


davidw:
I'm going to close this until the tests are fixed.

PR #43, opened 2014-12-22
Fixed tests for point, geometry and hstore types

hairyhum:
Added byte_size for point type encode.
Added update_type_cache for dynamic type tests.
Added point and geometry columns to test table


davidw:
This is fixed in the devel branch.

PR #47, opened 2015-01-23
Use 127.0.0.1 instead localhost to avoid IPv6 issues

matwey:
When localhost is used, it can be substituted by ::1
Our pg_hba.conf allows 127.0.0.1 only, so it leads to the following issue:

```
  psql: FATAL:  no pg_hba.conf entry for host "::1"
```

In order to avoid this, specify IPv4 addr explicitly


davidw:
Thanks!

PR #48, opened 2015-01-24
Encode undefined as NULL

tnt-dev:


davidw:
Could you provide some tests that show how this works, please?  All the tests pass with it, so I suppose it doesn't break anything, but it'd be nice to provide a test showing the effects of the patch.

tnt-dev:
I have added the appropriate tests.

davidw:
Thank you!

PR #51, opened 2015-02-10
int4range support

oferrigni:
This adds int4range as a supported wire type.


davidw:
Merged into the devel branch - thank you!

davidw:
Is there something that could be added to the README to document the presence of this type?

Issue #50, opened 2015-02-04
Add travis.ci support


seriyps:
Now, when we can setup and run tests fully automatically, it's good time to add CI support.

You may use, say, https://github.com/sfackler/rust-postgres/blob/master/.travis.yml as a reference.

Travis docs about postgresql setup: http://docs.travis-ci.com/user/using-postgresql/ (but we can just use our existing tests setup mechanism)

davidw:
Hrm; I wonder if that works out ok with the custom pg_hba.conf file we need?

seriyps:
@davidw see https://github.com/sfackler/rust-postgres/blob/master/.travis/setup.sh

seriyps:
@davidw but that example uses system postgresql. We can launch own instance, reusing existing test infrastructure.

davidw:
I'm happy to include this, but I'm afraid I'll probably need some help to get it off the ground.

seriyps:
Fixed by #53

PR #53, opened 2015-02-10
Travis support

oferrigni:
Travis will install postgres + contrib, build and execute tests.
Docs suggest waiting for a passing build in travis for PRs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,16 @@
+language: erlang
+otp_release:
+  - R16B03-1
+before_script:
+  - sudo apt-get update -qq
+  - sudo apt-get install postgresql-9.3-postgis-2.1 postgresql-contrib
+addons:
+  postgresql: "9.3"
+script:
+  - sudo chmod 777 /var/run/postgresql/
+  - ./setup_test_db.sh
+  - nohup ./start_test_db.sh &
+  - sleep 1
+  - make create_testdbs
+  - make test
+
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: erlang
 otp_release:
   - R16B03-1
+  - 17.4
 before_script:
   - sudo apt-get update -qq
   - sudo apt-get install postgresql-9.3-postgis-2.1 postgresql-contrib
@@ -9,7 +10,7 @@ addons:
 script:
   - sudo chmod 777 /var/run/postgresql/
   - ./setup_test_db.sh
-  - nohup ./start_test_db.sh &
+  - ./start_test_db.sh &
   - sleep 1
   - make create_testdbs
   - make test
--- a/README.md
+++ b/README.md
@@ -421,6 +421,7 @@ Here's how to create a patch that's easy to integrate:
 * Create a new branch for the proposed fix.
 * Make sure it includes a test and documentation, if appropriate.
 * Open a pull request against the `devel` branch of epgsql.
+* Passing build in travis
 
 ## Test Setup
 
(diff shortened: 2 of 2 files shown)


oferrigni:
Not trying to be opinionated. Didn't take long to do. Will need to setup at https://travis-ci.org/repositories to enable. Build passed in my fork @ https://travis-ci.org/oferrigni/epgsql-1 . README.md should point at https://travis-ci.org/epgsql/epgsql . Initial setup takes a bit, but runs pretty quick after that.

oferrigni:
Just saw that this might answer https://github.com/epgsql/epgsql/issues/50

seriyps:
Maybe add more OTP releases?

.travis.yml
```diff
@@ -0,0 +1,16 @@
...
+  - R16B03-1
+before_script:
+  - sudo apt-get update -qq
+  - sudo apt-get install postgresql-9.3-postgis-2.1 postgresql-contrib
+addons:
+  postgresql: "9.3"
+script:
+  - sudo chmod 777 /var/run/postgresql/
+  - ./setup_test_db.sh
+  - nohup ./start_test_db.sh &
```
seriyps: is `nohup` really required here?

oferrigni: i think start_test_db.sh blocks

oferrigni:
Did you have a list of otp versions you'd like? R16B03-1 is the minimum version and I didn't see any feedback that greater versions were being targeted on a roadmap.

.travis.yml
seriyps: yes, `start_test_db.sh` blocks, but I think just `&` may be enough. Anyway, this is not a principal question.

seriyps:
As a rule of thumb, every library tries to be compatible with last 3 major Erlang releases. For now it is R15, R16, 17. Not sure about minor releases, but why not?
Full releases list can be found there http://www.erlang.org/download.html

oferrigni:
epgsql does not compile in R15, I don't think.

seriyps:
Oh, yeah. Looks like it doesn't compile on R15 because `query` was reserved word before R16 (https://github.com/erlang/otp/commit/0dc3a29744bed0b7f483ad72e19773dc0982ea50). This word is used in typespec, so, compilation failed with

```
Compiling src/epgsql.erl failed:
src/epgsql.erl:30: syntax error before: '('
src/epgsql.erl:122: syntax error before: 'query'
src/epgsql.erl:140: syntax error before: 'query'
src/epgsql.erl:158: syntax error before: 'query'
ERROR: compile failed while processing /home/seriy/workspace/epgsql: rebar_abort
```
When I renamed `-type query()` to smth else, it compiles with no problems. I think this should be reported separately and fixed. @davidw what do you think?

davidw:
Yeah, that sounds like a separate issue that's worth fixing.

`start_test_db.sh` does block - it's kind of nice to be able to look at the queries going by.

oferrigni:
Per PR comments, I have removed the nohup (which now intermingles queries with eunit output giving some context to what is happening behind the scenes) and I added 17.4 as the most recent published release. 

Passing build @ https://travis-ci.org/oferrigni/epgsql-1/builds/50362253

oferrigni:
Are there other concerns or requested changes with this PR? I am worried I may have misunderstood some of the feedback. @seriyps @davidw

seriyps:
I think it's ok now. More erlang releases can be added when `query` issue will be fixed later.

davidw:
I merged it.... is there anything else we need to do to make it work?

oferrigni:
I think you need to go to https://travis-ci.org/epgsql/epgsql and enable the hook. Requires your authorizing travis to access the repo.

davidw:
Ok, I went and clicked some checkboxes on the Travis site.  We'll see if things start building...

seriyps:
There is also docs http://docs.travis-ci.com/user/getting-started/#Step-two%3A-Activate-GitHub-Webhook

davidw:
I can't tell if this is a problem with the Travis setup or just Travis catching a problem: https://github.com/epgsql/epgsql/pull/58

seriyps:
copy-paste my investigation here too:
> looks like R15B03-1 isn't available on travis and this is known problem: https://github.com/erlydtl/erlydtl/blob/master/.travis.yml#L11
> I think we should replace R15B03-1 withR15B03 on travis.yml

davidw:
Yep, looks good; done.

PR #55, opened 2015-02-13
Rename query keyword

seriyps:
Since `query` was reserved keyword prior R16, it can't be used in erlang code. I just renamed it to make epgsql compatible with R15 again. See https://github.com/epgsql/epgsql/pull/53#issuecomment-73803153 for details.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: erlang
 otp_release:
+  - R15B03-1
   - R16B03-1
   - 17.4
 before_script:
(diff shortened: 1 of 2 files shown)


oferrigni:
Looks good to me, fwiw.  Once Travis is on, we should see it pass across the board.

davidw:
Does this cause any backwards compatibility issues?  In other words, if someone were using Dialyzer, and relying on the existence of an epgsql:query type, it wouldn't work, correct?

That's not a problem as this is the right thing to do, it's just for the version number.

seriyps:
@davidw no, this type isn't exported (`-export_type`)

davidw:
Ok, cool, thanks!

PR #56, opened 2015-02-18
Properly cancel timed-out queries from SSL connections

pnc:
Formerly, the `cancel` handler tried to use `inet:peername/1` to determine where to connect to issue the cancel command. This caused a crash when the connection hosting the timed-out query was over SSL, since you must instead call `ssl:peername/1` for SSL sockets.

The crash looked like this:

    [error] CRASH REPORT Process <0.25743.31> with 2 neighbours exited
    with reason: no function clause matching
    prim_inet:peername({sslsocket,{gen_tcp,#Port<0.667326>,
                        tls_connection},<0.25810.31>})
    in gen_fsm:terminate/7 line 622

You can reproduce this by connecting to a server over SSL and issuing a `pg_sleep(120)` command.

This patch just changes the `cancel` handler to honor the `mod` atom already stored in the fsm's `State`.

### DIFF
--- a/src/epgsql_sock.erl
+++ b/src/epgsql_sock.erl
@@ -120,8 +120,12 @@ handle_cast({{Method, From, Ref}, Command} = Req, State)
 handle_cast(stop, State) ->
     {stop, normal, flush_queue(State, {error, closed})};
 
-handle_cast(cancel, State = #state{backend = {Pid, Key}}) ->
-    {ok, {Addr, Port}} = inet:peername(State#state.sock),
+handle_cast(cancel, State = #state{backend = {Pid, Key},
+                                   sock = TimedOutSock}) ->
+    {ok, {Addr, Port}} = case State#state.mod of
+                             gen_tcp -> inet:peername(TimedOutSock);
+                             ssl -> ssl:peername(TimedOutSock)
+                         end,
     SockOpts = [{active, false}, {packet, raw}, binary],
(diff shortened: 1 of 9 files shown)


pnc:
Looks like this build failed due to a missing `.travis.yml`? See https://github.com/epgsql/epgsql/pull/57.

seriyps:
You should rebase over 'devel' branch, not master.

src/epgsql_sock.erl
```diff
@@ -120,8 +120,12 @@ handle_cast({{Method, From, Ref}, Command} = Req, State)
...
-    {ok, {Addr, Port}} = inet:peername(State#state.sock),
+handle_cast(cancel, State = #state{backend = {Pid, Key},
+                                   sock = TimedOutSock}) ->
+    {ok, {Addr, Port}} = case State#state.mod of
+                             gen_tcp -> inet:peername(TimedOutSock);
+                             ssl -> ssl:peername(TimedOutSock)
+                         end,
     SockOpts = [{active, false}, {packet, raw}, binary],
     %% TODO timeout
     {ok, Sock} = gen_tcp:connect(Addr, Port, SockOpts),
```
seriyps: in case of `ssl` we anyway use `gen_tcp:connect/3`?

pnc:
Closing since this needs to be recreated against `devel`.

PR #58, opened 2015-02-18
Properly cancel timed-out queries from SSL connections

pnc:
Formerly, the `cancel` handler tried to use `inet:peername/1` to determine where to connect to issue the cancel command. This caused a crash when the connection hosting the timed-out query was over SSL, since you must instead call `ssl:peername/1` for SSL sockets.

The crash looked like this:

    [error] CRASH REPORT Process <0.25743.31> with 2 neighbours exited
    with reason: no function clause matching
    prim_inet:peername({sslsocket,{gen_tcp,#Port<0.667326>,
                        tls_connection},<0.25810.31>})
    in gen_fsm:terminate/7 line 622

You can reproduce this by connecting to a server over SSL and issuing a `pg_sleep(120)` command.

This patch just changes the `cancel` handler to honor the `mod` atom already stored in the fsm's `State`.

This leaves the code as it is so it connects over regular TCP even for SSL connections, since Postgres just wants the secret key to match. [The manual](http://www.postgresql.org/docs/9.3/static/protocol-flow.html#AEN99368) doesn't mention whether SSL is required for cancelation. My experiments indicate it isn't, but I can add support for upgrading to SSL before issuing the cancel if it seems necessary.

### DIFF
--- a/src/epgsql_sock.erl
+++ b/src/epgsql_sock.erl
@@ -120,8 +120,12 @@ handle_cast({{Method, From, Ref}, Command} = Req, State)
 handle_cast(stop, State) ->
     {stop, normal, flush_queue(State, {error, closed})};
 
-handle_cast(cancel, State = #state{backend = {Pid, Key}}) ->
-    {ok, {Addr, Port}} = inet:peername(State#state.sock),
+handle_cast(cancel, State = #state{backend = {Pid, Key},
+                                   sock = TimedOutSock}) ->
+    {ok, {Addr, Port}} = case State#state.mod of
+                             gen_tcp -> inet:peername(TimedOutSock);
+                             ssl -> ssl:peername(TimedOutSock)
+                         end,
     SockOpts = [{active, false}, {packet, raw}, binary],
     %% TODO timeout
     {ok, Sock} = gen_tcp:connect(Addr, Port, SockOpts),


src/epgsql_sock.erl
```diff
@@ -120,8 +120,12 @@ handle_cast({{Method, From, Ref}, Command} = Req, State)
...
-    {ok, {Addr, Port}} = inet:peername(State#state.sock),
+handle_cast(cancel, State = #state{backend = {Pid, Key},
+                                   sock = TimedOutSock}) ->
+    {ok, {Addr, Port}} = case State#state.mod of
+                             gen_tcp -> inet:peername(TimedOutSock);
+                             ssl -> ssl:peername(TimedOutSock)
+                         end,
     SockOpts = [{active, false}, {packet, raw}, binary],
     %% TODO timeout
     {ok, Sock} = gen_tcp:connect(Addr, Port, SockOpts),
```
seriyps: Again, is it ok to connect via `gen_tcp` in case, when `#state.mod` is `ssl`?

davidw:
I can't tell whether the Travis error is legit or the result of an improper setup, since it's a recent addition.

seriyps:
@davidw  Looks like travis can't find R15B3-1 https://travis-ci.org/epgsql/epgsql/jobs/51234524#L92

seriyps:
@davidw looks like R15B03-1 isn't available on travis and this is known problem: https://github.com/erlydtl/erlydtl/blob/master/.travis.yml#L11
I think we should replace `R15B03-1` with` R15B03` on travis.yml

src/epgsql_sock.erl
pnc: In my tests, Postgres accepted cancel commands over TCP even if they were canceling a query that had been issued over SSL. [The manual](http://www.postgresql.org/docs/9.3/static/protocol-flow.html#AEN99368) doesn't seem to cover what to do here. As I read it, [libpq doesn't use SSL either](https://github.com/postgres/postgres/blob/080eabe2e8a184ff40b7380aaaa9418714acace9/src/interfaces/libpq/fe-connect.c#L3240). Thoughts?

pnc:
@seriyps Let me know when the travis.yml is changed and I'll rebase this so it builds again.

davidw:
Travis is fixed, but there's an error completely unrelated to this on R15, since I don't think anyone has tested it there.  We just introduced Travis, so please bear with us as we iron out the kinks.

src/epgsql_sock.erl
davidw: It might be worth asking on one of the Postgres lists to ensure we get it right.  I asked around a bit on #postgresql and there were no strong opinions one way or the other.

seriyps: Looks like reference implementation (libpq) has API `PQcancel` for this http://www.postgresql.org/docs/8.4/static/libpq-cancel.html (at least, pgadmin3 uses it http://git.postgresql.org/gitweb/?p=pgadmin3.git;a=blob;f=pgadmin/db/pgConn.cpp;h=c694d1e93da7eafd99d8ea96dc01e1aa5b79247e;hb=HEAD#l1050)

And there is it's implementation http://git.postgresql.org/gitweb/?p=postgresql.git;a=blob;f=src/interfaces/libpq/fe-connect.c;h=25961b1f10dc59b1aa86630068b8e1ab1be0d8ff;hb=HEAD#l3217

seriyps: So, the short answer is "no, current code is ok and no ssl needed"

davidw:
Cool, thank you!

src/epgsql_sock.erl
pnc: Thanks for the second pair of eyes on libpq! :+1:

seriyps: @pnc I'm actualy just doesn't noticed, that you added link to libpq sources "libpq doesn't use SSL either". That's why I started my own investigation =)

PR #63, opened 2015-05-01
epgsqla:equery check params

jamhed:
When parameter types and actual value types mismatches the socket process crashes
closing connection to the server. Instead of crash this patch returns a message to sender
{parse_error, Ref, Reason}.


davidw:
Should be against the devel branch, please.  However, people may have written code that expects things to crash if a bad query comes in, so I'm unsure about this.

PR #65, opened 2015-06-01
add #error.codename field

RJ:
`epgsql_errcodes.erl` is generated from the postgres source, using the script added to `scripts/`, which does lots of this sort of thing:

    to_name(<<"23502">>) -> not_null_violation;
    to_name(<<"23503">>) -> foreign_key_violation;
    to_name(<<"23505">>) -> unique_violation;
    to_name(<<"23514">>) -> check_violation;
    ... etc ...

Added a codename field to the #error{} record.

Now you can match epgsql errors like this:

    case epgsql:something() of
      {error, #error{codename = foreign_key_violation}} -> ...
    end

You can still match `#error{code = <<"23503">>}` too.

I bumped the minor version - might warrant a major version bump, since it changes the #error{} field in the public header (everything needs recompiling). I think it's worthwhile though - matching on error names is so much nicer.


davidw:
Hi,

Could you point this at the devel branch, rather than master?

RJ:
see #66, which points at devel not master

PR #66, opened 2015-06-02
add #error.codename field

RJ:
`epgsql_errcodes.erl` is generated from the postgres source, using `generate_errcodes_src.sh`, which does lots of this sort of thing:

    to_name(<<"23502">>) -> not_null_violation;
    to_name(<<"23503">>) -> foreign_key_violation;
    to_name(<<"23505">>) -> unique_violation;
    to_name(<<"23514">>) -> check_violation;
    ... etc ...

Added a codename field to the #error{} record.

Now you can match epgsql errors like this:

    case epgsql:something() of
      {error, #error{codename = foreign_key_violation}} -> ...
    end

You can still match `#error{code = <<"23503">>}` too.

I bumped the minor version - might warrant a major version bump, since it changes the #error{} field in the public header (everything needs recompiling). I think it's worthwhile though - matching on error names is so much nicer.


seriyps:
Looks like Travis tests failed.

scripts/generate.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
```
seriyps: Maybe use more verbose script name? `generate.sh` has not so much meaning.

seriyps:
Yeah, looks like build failed because in unittests record match defined not like record, but like regular tuple:

https://github.com/RJ/epgsql/blob/errcodes/test/epgsql_tests.erl#L24

davidw:
Looks good to me - any other comments, @seriyps ?

seriyps:
@davidw no. Looks good.

davidw:
Thanks for this, @RJ !

RJ:
thanks!

PR #68, opened 2015-08-03
dialyzer sends a warning when using epgsql

Littlelies:
squery_row type definition is too restrictive. If you retrieve multiple columns, you don't get a {binary()}, but a tuple().


seriyps:
Most likely, `{binary()}` was wrongly interpreted as 'tuple of binaries', like analogy of `[binary()]` list of binaries.
So, looks good.

davidw:
@Littlelies could I trouble you to redo this against the devel branch?  Thank you

seriyps:
Also, `equery_row` should be changed the same way.

seriyps:
Fixed by #105

PR #72, opened 2015-09-27
Work around some dependent types. Fixes dialyzer runs.

### LINKED ISSUE #71: Dialyzer errors in include/epgsql_geometry.hrl
Hello,

For some reason I am getting dialyzer errors after upgrading to OTP 18.1 (from OTP 17):

```
$ rebar build-plt
ERROR: Dialyzer error:
Analysis failed with error:
epgsql_geometry.hrl:97: Illegal declaration of #multi_surface{point_type}
```

This seems weird, considering that there has been a commit that was supposed to fix a lot of dialyzer errors. Am I using an incorrect version of something? What can I do to find out what is causing this error?

```
$ dialyzer --version
Dialyzer version v2.8.1
$ erl --version
Erlang/OTP 18 [erts-7.1] [source] [64-bit] [smp:4:4] [async-threads:10] [hipe] [kernel-poll:false]
```

The interesting thing is that if I revert the commit that was supposed to fix dialyzer errors, the issue goes away. Which confuses me even more.

solatis:
Fixes #71

### DIFF
--- a/include/epgsql_geometry.hrl
+++ b/include/epgsql_geometry.hrl
@@ -1,7 +1,7 @@
 -type point_type() :: '2d' | '3d' | '2dm' | '3dm'.
 
 -record(point,{
-  point_type :: point_type(),
+  point_type :: any(),
   x :: float(),
   y :: float(),
   z :: float(),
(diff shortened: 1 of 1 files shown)


davidw:
It's best to do pull requests against the devel branch rather than master.  Thanks, though!

PR #74, opened 2015-10-20
Fix dialyzer for Erlang 18 compat

seth:


davidw:
Hi, this looks good - does it cause any problems for older versions? Also, could you do the pull request against `devel` rather than master?  Thank you!

seth:
Oops, sent this to the wrong remote.

PR #76, opened 2015-10-21
Allow type cache query to fail

edhollandAL:
For version of PG below 9 the pg_type table has a different format this causes epgsql to fail to connect to AWS redshift as it is based on PG 8

### DIFF
--- a/src/epgsql.erl
+++ b/src/epgsql.erl
@@ -109,8 +109,15 @@ update_type_cache(C) ->
     Query = "SELECT typname, oid::int4, typarray::int4"
             " FROM pg_type"
             " WHERE typname = ANY($1::varchar[])",
-    {ok, _, TypeInfos} = equery(C, Query, [DynamicTypes]),
-    ok = gen_server:call(C, {update_type_cache, TypeInfos}).
+    case equery(C, Query, [DynamicTypes]) of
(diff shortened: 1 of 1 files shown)


src/epgsql.erl
```diff
@@ -109,8 +109,14 @@ update_type_cache(C) ->
     Query = "SELECT typname, oid::int4, typarray::int4"
             " FROM pg_type"
             " WHERE typname = ANY($1::varchar[])",
-    {ok, _, TypeInfos} = equery(C, Query, [DynamicTypes]),
-    ok = gen_server:call(C, {update_type_cache, TypeInfos}).
+    case equery(C, Query, [DynamicTypes]) of
+        {ok, _, TypeInfos} ->
+            ok = gen_server:call(C, {update_type_cache, TypeInfos});
+        {error, _Error} ->
```
seriyps: Match with some concrete `_Error` would be better. For example, what if it's socket error or like that?

edhollandAL:
Have updated to only silently ignore the specific error we encounter

edhollandAL:
@seriyps: Have updated based on your comment. Are you happy with this now?

seriyps:
yep, looks better.

edhollandAL:
@davidw: are we ready to merge?

PR #81, opened 2015-12-08
support custom types in prepared statements

id:
so that one can call e.g. epgsql:update_type_cache(C, [<<"my_enum1">>, <<"my_enum2">>])


davidw:
This needs to be done against the devel branch.

PR #83, opened 2015-12-09
support parsed statements in epgsql:equery

id:
add a clause for epgsql:equery/3 which accepts parsed statements and
don't require parse

### DIFF
--- a/src/epgsql.erl
+++ b/src/epgsql.erl
@@ -135,7 +135,9 @@ squery(Connection, SqlQuery) ->
 equery(C, Sql) ->
     equery(C, Sql, []).
 
-%% TODO add fast_equery command that doesn't need parsed statement
+equery(C, #statement{types = Types} = S, Parameters) ->
+    Typed_Parameters = lists:zip(Types, Parameters),
+    gen_server:call(C, {equery, S, Typed_Parameters}, infinity);
 equery(C, Sql, Parameters) ->
     case parse(C, "", Sql, []) of
         {ok, #statement{types = Types} = S} ->


id:
Does this change make sense? If so, I'll add a test and update README.

id:
This won't work because equery command in epgsql_sock closes the statement.

So basically I need 'execute_batch' but for a single statement. Do you think it can be a separate api?

davidw:
@id I don't work with this on a day to day basis, so a bit more explanation of what you're trying to do would be really helpful.

id:
@davidw I don't see an option to prepare a statement and then do a blocking call to epgsql process providing the statement and parameters. With equery there is one more 'parse' call which I can avoid.

I could use e.g. 'execute_batch' providing my statement with parameters as a list with single item, but for some reason it only returns values and not columns (I need columns).

For example:

```
epgsql:squery(C, "create table foobar (x int, y text);").
epgsql:squery(C, "insert into foobar (x, y) values (1, '1');").
```

This is what I get from equery:

````
> epgsql:equery(C, "select * from foobar where x = $1;", [1]).
{ok,[{column,<<"x">>,int4,4,-1,1},
     {column,<<"y">>,text,-1,-1,1}],
    [{1,<<"1">>}]}
```

And this is what I get from execute_batch:

```
{ok, S} = epgsql:parse(C, "select_foobar", "select * from foobar where x = $1;", [int4]).
> epgsql:execute_batch(C, [{S, [1]}]).
[{ok,[{1,<<"1">>}]}]
```

Not really sure what's the reason for the difference in behaviour.

Now there is another problem which I found while looking around.
epgsqla:equery/3 claims to accept parsed statements while in fact you can only call epgsqla:equery/3 once for one statement. Next time one would have to 'parse' the statement again, otherwise it will crash because epgsql_sock closes the statement in 'equery' command.

One more issue is that epgsqla:equery/2 takes plain Sql according to argument name :)

I would like to have a function which works like equery, but takes already parsed statement as an argument and doesn't close the statement. What do you think about this?

id:
@davidw ping

davidw:
@seriyps did you have a look at this?

seriyps:
I'm not sure if this is different from `parse` (once), then `bind` and `execute` multiple times.

Anyway, some tests required for suggested changes.

id:
Sorry guys, it seems like this one is not a good idea.

I've experienced strange postgres behaviour while having a couple of opened parsed statements on several connections. It was like some of database operations were blocked for other connections.

Don't have time to investigate this now.

PR #84, opened 2015-12-10
close socket in terminate/2

id:
in some case when a user process got timeout and reconnected,
an old connection to postgres might stay hanging around for quite a while

### DIFF
--- a/src/epgsql_sock.erl
+++ b/src/epgsql_sock.erl
@@ -152,11 +152,15 @@ handle_info({inet_reply, _, Status}, State) ->
 handle_info({_, Sock, Data2}, #state{data = Data, sock = Sock} = State) ->
     loop(State#state{data = <<Data/binary, Data2/binary>>}).
 
-terminate(_Reason, #state{sock = Sock}) when is_port(Sock) ->
-    gen_tcp:close(Sock),
-    ok;
-terminate(_Reason, _State) ->
-    ok.
+terminate(_Reason, #state{mod = Mod, sock = Sock}) ->
+    case {Mod, Sock} of
+      {_, undefined} ->
+        ok;
+      {gen_tcp, _} ->
+        gen_tcp:close(Sock);
+      {ssl, _} ->
+        ssl:close(Sock)
+    end.
 
 code_change(_OldVsn, State, _Extra) ->
     {ok, State}.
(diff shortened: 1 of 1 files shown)


davidw:
I realize this is a tricky one to test, but the more tests, the better.

src/epgsql_sock.erl
```diff
@@ -152,8 +152,10 @@ handle_info({inet_reply, _, Status}, State) ->
 handle_info({_, Sock, Data2}, #state{data = Data, sock = Sock} = State) ->
     loop(State#state{data = <<Data/binary, Data2/binary>>}).
 
+terminate(_Reason, #state{sock = Sock}) when is_port(Sock) ->
+    gen_tcp:close(Sock),
```
seriyps: And what about ssl?

id: Thanks, fixed.

src/epgsql_sock.erl
```diff
@@ -152,9 +152,15 @@ handle_info({inet_reply, _, Status}, State) ->
 handle_info({_, Sock, Data2}, #state{data = Data, sock = Sock} = State) ->
     loop(State#state{data = <<Data/binary, Data2/binary>>}).
 
-terminate(_Reason, _State) ->
-    %% TODO send termination msg, close socket ??
-    ok.
+terminate(_Reason, #state{mod = Mod, sock = Sock}) ->
```
seriyps: why not smth like

```
terminate(_, #state{sock=undefined}) -> ok;
terminate(_, #state{mod=gen_tcp}) -> gen_tcp:close(...);
terminate(..., ..mod=ssl}) -> ssl:close(...).
```
:smile:

neeraj9:
Any specific reason this pull still open? According to me this is an important change.

seriyps:
I'm ok with this.
@davidw ?

The only question is: AFAIK, when process, holding a connection dies, socket will be closed anyway. So, why implicitly close it?
Or, when process dies, socket closed in some other way, compared to explicit close?

id:
Socket is not always closed immediately, it could stay open on OS level for a while even after your application restarted. Then the application will re-establish connection to postgres and it could lead to the situation when you exhausted limit on available connections for your user.

seriyps:
Ok, let's wait @davidw for some time. For me it's ok.

seriyps:
Thank you for patience and your work.

davidw:
Sorry for the late reply but I wanted to say it looks like a good idea, and thanks!

neeraj9:
Thanks. Saves me the trouble of keeping private merges.

PR #85, opened 2015-12-16
Fixing error in comments.

aherranz:
I think the documentation (and comments in code) are wrong. I have also added delete statement to the comments (with insert and update).


seriyps:
Looks good.

davidw:
Pull requests, as per the documentation, should be made against devel. Thanks!

seriyps:
May you also, please, update the same comments in README?

aherranz:
From devel (sorry), and README updated: #87

PR #87, opened 2015-12-16
Fixing error in comments, added delete statement to the

aherranz:
Fixing error in comments, added delete statement to the comments (with insert and update). Progressed in README.


davidw:
Thanks!

PR #88, opened 2016-01-01
epgsql_sock: aggregate small packets into one

yjh0502:
Aggregate small packets into large one to reduce number of
port_command/2 calls, and eventually reduce number of packets.

### DIFF
--- a/src/epgsql_sock.erl
+++ b/src/epgsql_sock.erl
@@ -217,17 +217,21 @@ command({equery, Statement, Parameters}, #state{codec = Codec} = State) ->
     #statement{name = StatementName, columns = Columns} = Statement,
     Bin1 = epgsql_wire:encode_parameters(Parameters, Codec),
     Bin2 = epgsql_wire:encode_formats(Columns),
-    send(State, ?BIND, ["", 0, StatementName, 0, Bin1, Bin2]),
-    send(State, ?EXECUTE, ["", 0, <<0:?int32>>]),
-    send(State, ?CLOSE, [?PREPARED_STATEMENT, StatementName, 0]),
-    send(State, ?SYNC, []),
+    send_multi(State, [
+        {?BIND, ["", 0, StatementName, 0, Bin1, Bin2]},
+        {?EXECUTE, ["", 0, <<0:?int32>>]},
+        {?CLOSE, [?PREPARED_STATEMENT, StatementName, 0]},
+        {?SYNC, []}
+    ]),
     {noreply, State};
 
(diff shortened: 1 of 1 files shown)


davidw:
Could this have any adverse affects if one of the commands in the middle fails?

yjh0502:
There's no middles fail in this case. It's about sending bytes to server, and in TCP application should work same if packets are representing same byte stream. Some poorly-written servers assume about how client send packets, but I believe PostgreSQL is not the case.

davidw:
Thanks! Out of curiosity do you have any benchmarks?

src/epgsql_sock.erl
```diff
@@ -235,43 +239,49 @@ command({bind, Statement, PortalName, Parameters}, #state{codec = Codec} = State
...
     {noreply, State};
 
-command({execute_batch, Batch}, State) ->
-    #state{mod = Mod, sock = Sock, codec = Codec} = State,
-    BindExecute =
-        lists:map(
-          fun({Statement, Parameters}) ->
+command({execute_batch, Batch}, #state{codec = Codec} = State) ->
+    Commands =
+        lists:foldr(
```
seriyps: Why not map?

yjh0502:
@davidw Here's some benchmark [1] numbers [2]

queue length | queries per second | queries per second with patch | improvement
---- | ---- | ---- | ----
5 | 6,716.61 | 9,483.44 | 141.19%
10 | 7,070.47 | 11,002.74 | 155.62%
20 | 7,377.38 | 12,008.60 | 162.78%
50 | 6,068.76 | 10,915.24 | 179.86%
100 | 6,062.78 | 11,182.67 | 184.45%

Benchmark Environment
 - PostgreSQL 9.3
 - AMD Phenom(tm) II X4 945
 - Both client and server resides on same machine

The benchmark uses `epgsqla` interface to run queries, and measures how many queries per second `epgsql` can handle while maintaining given number of queries on queue.

[1] https://gist.github.com/yjh0502/ecb9c899fbbb020b4e33
[2] https://docs.google.com/spreadsheets/d/1n_tZ3ypmZlqIoxLL2eWqSRuuMthxhdE_PBBloBVhLcY/edit?usp=sharing

davidw:
Well that's nice work indeed! Thank you!

sdebnath:
Was this tagged on git? Would love to test it out. Thanks.

davidw:
@sdebnath it's in the devel branch and will be in the next release, whenever that happens.

PR #92, opened 2016-02-03
add prepared queries function

eshubin:
With prepared query it's possible to create a named template of query with epgsql:parse on application start and query it afterwards with different arguments.


seriyps:
Looks interesting.

May you also add some unittests for success and error situations?

eshubin:
@seriyps here you are.

seriyps:
@davidw ?

davidw:
So basically a convenience method that lets us avoid doing the whole describe/bind/execute dance?

I'm not sure I like those nested case statements... maybe it should just crash if one does not return ok?

fishcakez:
It would be more efficient to follow exactly the same pattern as `equery` except only to describe and not parse in the first step.

fishcakez:
It might also be useful to have an exported function that does the second call of `equery` so the statement record can be reused and a prepared query only requires a single round trip to the database.

eshubin:
@davidw I think I addressed nested cases, originally did not want to go deep into epgsql_socket module.
@fishcakez is the last commit what you mean?

fishcakez:
> @fishcakez is the last commit what you mean?

In `devel` there is a multi_send function which look more efficient than multiple send calls, you may also want to use that - `equery` uses that in `devel`. But that aside yes I think so. I'm not familiar with `epsql`.

eshubin:
@fishcakez @davidw @seriyps I think I've addressed all comments in the pool request. Are there any further actions required from my side?

davidw:
Thanks, @eshubin !

PR #93, opened 2016-03-03
epgsqla: add prepared_query/3

yjh0502:


seriyps:
Please, add:

* Unittests
* Typespec
* Examples in README

yjh0502:
@seriyps 

I added
 - unit tests for epgsql{a,i}:prepared_query/3
 - typespecs for epgsql{a,i}:prepared_query/3. Typespecs for epgsql{a,i}:equery/4 was incorrect so I fixed it.
 - examples on README, also explanation about typed parameters on README.

src/epgsqla.erl
```diff
@@ -55,9 +56,13 @@ squery(C, Sql) ->
 equery(C, Sql) ->
     equery(C, Sql, []).
 
--spec equery(epgsql:connection(), #statement{}, [epgsql:bind_param()]) -> reference().
-equery(C, Statement, Parameters) ->
-    cast(C, {equery, Statement, Parameters}).
+-spec equery(epgsql:connection(), #statement{}, [epgsql:typed_param()]) -> reference().
+equery(C, Statement, TypedParameters) ->
+    cast(C, {equery, Statement, TypedParameters}).
```
seriyps: maybe we should move lists:zip inside this function? Or this will break backward compatibility?

seriyps:
Looks good!

@davidw any thoughts?

davidw:
I would maybe add just a bit more explaining what this is useful for.

yjh0502:
@davidw @seriyps 
I just updated benchmark[1] numbers[2] with `epgsqla:prepared_query/3`.


queue length | queries per second | queries per second with patch | improvement
---- | ---- | ---- | ----
5 | 9,483 | 23,285 | 246%
10 | 11,003 | 26,204 | 238%
20 | 12,009 | 28,300 | 236%
50 | 10,915 | 25,178 | 231%
100 | 11,183 | 23,514 | 210%

In this case, performance gain is came by eliminating database round-trip with `describe`/`parse` command. If the prepared query is complex enough, we can expect additional gain from skipping query parsing/planning for each execution.

Some thoughts

 - `epgsql:prepared_query/3` sends `describe` command to server for each `prepared_query/3` call, so it still requires two round-trips for each call. We can eliminate the overhead by taking `#statement{}` as an argument instead of query name.
 - As @seriyps pointed out, we can move `lists:zip(Types, Parameters)` to `epgsql_sock:command/2`. It will break compatibility, but I think it's cleaner/more intuitive to users.

[1] https://gist.github.com/yjh0502/ecb9c899fbbb020b4e33#gistcomment-1714518
[2] https://docs.google.com/spreadsheets/d/1n_tZ3ypmZlqIoxLL2eWqSRuuMthxhdE_PBBloBVhLcY

seriyps:
@davidw motivation is the same as for https://github.com/epgsql/epgsql/pull/92 : parse query once and then repeat this query multiple times with different placeholder parameters through the same connection.
And this patch just add this `epgsql:prepared_query` to `epgsqla` and `epgsqli` at least for consistency.

@yjh0502 regarding `epgsql:prepared_query/3`: maybe add 2'nd clause (physically it will be first clause), which accepts `#statement{}` record? But in separate PR please =)
About `TypedParameters`: I think we shouldn't break backward compatibility. At least until some next major release. Maybe separate issue for this?

davidw:
I agree about not breaking backwards compatibility.

yjh0502:
@seriyps @davidw I'm okay for not breaking backwards compatibility.

seriyps:
@davidw may we merge this? It doesn't break backward compatibility and makes epgsql API consistent across all 3 implementations (epgsql, epgsqla, epgsqli).

davidw:
Thanks!

Issue #14, opened 2014-03-28
json support


davidw:
https://github.com/soundrop/epgsql/commit/3afda8b7200805933e7d080cc293bde24f992349 looks like a first step.

laheadle:
@talentdeficit has a branch with tests! Is this mergeable?

https://github.com/talentdeficit/epgsql/commit/a09030d8c37edc0b70278cc88a213176411266d4

seriyps:
@laheadle looks good. But we need also fix travis.yml to use PG 9.4 (9.3 doesn't have jsonb support).

dxtr:
Has there been any progress on this?

seriyps:
@davidw what do you think? May we merge patch from https://github.com/talentdeficit/epgsql/commit/a09030d8c37edc0b70278cc88a213176411266d4 ? It looks good.

davidw:
Fixed, thanks to @talentdeficit and @seriyps - thanks guys!

PR #64, opened 2015-05-13
support for postgresql json type

talentdeficit:
support for the `json` type in postgres 9.2+

PR #94, opened 2016-04-19
support for postgresql json/jsonb type

seriyps:
This one superseeds #64, since it contains both json and jsonb and has no extra patches (like mention of unrelevant pgapp in README).

It's rebased over 'devel' patch from https://github.com/epgsql/epgsql/compare/master...talentdeficit:jsonb


seriyps:
So, finaly, working json & jsonb types support.

davidw:
Thanks, @seriyps and @talentdeficit !

seriyps:
Thanks! This also fixes #14 and #64

PR #96, opened 2016-05-19
added srid feature

namjae:
epgsql crashed with following message when i tried to access(decode) geometry column with SRID.
i added srid feature. i think this is a quick and dirty patch. but, in my case it works.
```
** exception exit: {{function_clause,
                        [{ewkb,decode_point_type,
                             [<<0,32>>],
                             [{file,"src/ewkb.erl"},{line,249}]},
                         {ewkb,decode_geometry_data,1,
                             [{file,"src/ewkb.erl"},{line,111}]},
                         {ewkb,decode_geometry,1,
                             [{file,"src/ewkb.erl"},{line,27}]},
                         {epgsql_wire,decode_data,4,
                             [{file,"src/epgsql_wire.erl"},{line,105}]},
                         {epgsql_sock,on_message,2,
                             [{file,"src/epgsql_sock.erl"},{line,610}]},
                         {epgsql_sock,loop,1,
                             [{file,"src/epgsql_sock.erl"},{line,334}]},
                         {gen_server,try_dispatch,4,
                             [{file,"gen_server.erl"},{line,615}]},
                         {gen_server,handle_msg,5,
                             [{file,"gen_server.erl"},{line,681}]}]},
                    {gen_server,call,
                        [<0.1603.0>,
                         {execute,tbl_gas_station_fetch_by_distance,
                             [126.880284585036,37.456134032755,100]},
                         infinity]}}
     in function  gen_server:call/3 (gen_server.erl, line 212)
     in call from sqerl:with_db/2 (src/sqerl.erl, line 71)
     in call from sqerl:execute_statement/4 (src/sqerl.erl, line 126)
     in call from sqerl:select/4 (src/sqerl.erl, line 93)
     in call from sqerl_rec:qfetch/3 (src/sqerl_rec.erl, line 132)
(mz_fuel_cp@127.0.0.1)7> 17:06:10.256 [error] gen_server <0.1604.0> terminated with reason: no function clause matching ewkb:decode_point_type(<<0,32>>) line 249
(mz_fuel_cp@127.0.0.1)7> 17:06:10.256 [error] CRASH REPORT Process <0.1604.0> with 1 neighbours exited with reason: no function clause matching ewkb:decode_point_type(<<0,32>>) line 249 in gen_server:terminate/7 line 826
[...]

PR #97, opened 2016-05-19
added srid feature

namjae:
moved from https://github.com/epgsql/epgsql/pull/96

### DIFF
--- a/src/ewkb.erl
+++ b/src/ewkb.erl
@@ -33,6 +33,11 @@ encode_geometry(Geometry) ->
   Data = encode_geometry_data(Geometry),
   <<1, Type/binary, PointType/binary, Data/binary>>.
 
+encode_geometry_data(#point{ point_type = '2d_srid', x = X, y = Y, srid = SRID }) ->
+  Xbin = encode_float64(X),
+  Ybin = encode_float64(Y),
+  SRIDbin = encode_int32(SRID),
+  <<SRIDbin/binary, Xbin/binary, Ybin/binary>>;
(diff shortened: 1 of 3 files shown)


seriyps:
Can you please add tests which actualy USE this new feature? So, `srid` field isn't `undefined`.

seriyps:
@namjae sorry, didn't get any notification about your update.
I'm mean that we need a test which actualy inserts and selects this `2d_srid` points to/from database.
Like `check_type(Module, geometry, "SELECT blablabla", {point, '2d_srid', blabla})`

src/ewkb.erl
```diff
@@ -176,6 +181,10 @@ decode_float64(<<Hex:8/binary, Rest/binary>>) ->
   <<Float:1/little-float-unit:64>> = Hex,
   {Float, Rest}.
 
+decode_point(PointType='2d_srid', Data) ->
+  {Value, R} = decode_int32(Data),
```
seriyps: I'd better sugest name `SRID` for this variable instead of `Value`.

seriyps:
I'm also wonder if we really need this special datatype for points with/without SRID. As far as I understand, this SRID may be attached to any type of point? Am I right? Maybe we should use original '2d' '3d' etc types, but distinguish them by just checking if `srid` record's field is `undefined`?

namjae:
hmm, i was making some prototype application back then.
and got some errors related with SRID.
so, i made a quick and dirty patch to solve the problem.
and i totally forgot about this PR.

srid is NULL (undefined in erlang) if not specified i remember.

and i had some problems with current/provided unit tests. i couldn't make all of them pass.

PR #98, opened 2016-08-04
Add `set_notice_receiver/2` function

seriyps:
Now postgresql notification (warnings and LISTEN/NOTIFY) receiver process can be configured only at connection start time and can be only of `pid()` type (via `{async, pid()}` `connect/4` option).

But in nowadays we usualy use some connection pools, which usualy configured declaratively (eg, via sys.config), and we can't say them "hey, use this pid for notifications" in config-file. So, to figure out this problem, possibility to pass atom as `async` option value was introduced.

Also, when using connection polling we may want to start receive notifications/notices by the process, which take this connection from the pool and reset this option before passing connection back. `set_notice_receiver/2` adresses exactly this issue.


seriyps:
https://github.com/epgsql/epgsql/pull/99

PR #99, opened 2016-08-04
Add `set_notice_receiver/2` function

seriyps:
(sorry, now over `devel` branch)
Now postgresql notification (warnings and LISTEN/NOTIFY) receiver process can be configured only at connection start time and can be only of pid() type (via {async, pid()} connect/4 option).

But in nowadays we usualy use some connection pools, which usualy configured declaratively (eg, via sys.config), and we can't say them "hey, use this pid for notifications" in config-file. So, to figure out this problem, possibility to pass atom as async option value was introduced.

Also, when using connection polling we may want to start receive notifications/notices by the process, which take this connection from the pool and reset this option before passing connection back. set_notice_receiver/2 adresses exactly this issue.


davidw:
You might make a small note in the docs about usage of this.

Looks good though, once tests pass.

seriyps:
@davidw do you think update in README isn't enough?

davidw:
@seriyps it's not necessary, but it'd be "nice to have" something that points out what this is for and how it might be used.  That could well just be a link to the Postgres docs.  To me it's sort of a little-known Postgres feature that I'm not sure is a very good idea to use.  Perhaps a use case or something would help people know to use it - or avoid it.

seriyps:
Links to docs added; also some extra tests; `#error.severity` clarified

seriyps:
Hey, @davidw. Anything to add?

PR #101, opened 2016-09-15
Feature/registered process

marcelloceschia:
allow async to be a atom, so it is posible to send async messages to a registered process

### DIFF
--- a/src/epgsql_sock.erl
+++ b/src/epgsql_sock.erl
@@ -426,11 +426,21 @@ notify(State = #state{queue = Q}, Notice) ->
     end,
     State.
 
-notify_async(State = #state{async = Pid}, Msg) when is_pid(Pid); is_atom(Pid) ->
-    Pid ! {epgsql, self(), Msg},
-    State;
+notify_process(undefined, Msg) ->
+    false;
 
-notify_async(State, _Msg) ->
+notify_process(Pid, Msg) when is_pid(Pid) ->
+    Pid ! {epgsql, self(), Msg},
+    true;
+    
+notify_process(Process, Msg) when is_atom(Process) ->
+    notify_process(whereis(Process), Msg);
+
+notify_process(_, _) ->
+    false.
+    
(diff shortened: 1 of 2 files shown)


seriyps:
https://github.com/epgsql/epgsql/pull/99

src/epgsql_sock.erl
```diff
@@ -426,11 +426,11 @@ notify(State = #state{queue = Q}, Notice) ->
     end,
     State.
 
-notify_async(State = #state{async = Pid}, Msg) ->
-    case is_pid(Pid) of
-        true  -> Pid ! {epgsql, self(), Msg};
-        false -> false
-    end,
+notify_async(State = #state{async = Pid}, Msg) when is_pid(Pid); is_atom(Pid) ->
+    Pid ! {epgsql, self(), Msg},
```
seriyps: It even will try to send messages to `undefined` and fail of course (if there is no process registered under `undefined` atom)

marcelloceschia: this issue was fixed in the latest commit

marcelloceschia: @seriyps somthing I need to change?

seriyps:
Implementation looks good. But there are 2 things extra:

* Unittests
* There is another pull-request which adds exactly the same functionality: #99 

We are waiting for @davidw approval.

marcelloceschia:
@seriyps you are right, #99 does nearly the same, sorry for reimplementing this again. But checking this code, it has the same issue you mentioned to me - there is no check if the named process is registered

seriyps:
@marcelloceschia there is no check, but there is a `try ... catch` (`atom() ! any()` will raise exception if no process registered). I decided that it will be optimal fast-path. Since passing atom with no process registered on it is more like error and shouldn't happen. I was thinking about adding some logging in 'catch' clause, but epgsql don't have any logging at all, so, not sure if I should add.

seriyps:
#99 with the same functionality was merged to `devel` branch. Thank you anyway!

PR #104, opened 2016-11-16
Make it compatible with Erlang 19. Add travis

seriyps:
Just make it work on Erlang 19

### DIFF
--- a/src/epgsql_wire.erl
+++ b/src/epgsql_wire.erl
@@ -209,6 +209,6 @@ encode_list(L) ->
     <<(byte_size(Bin)):?int32, Bin/binary>>.
 
 encode_standby_status_update(ReceivedLSN, FlushedLSN, AppliedLSN) ->
-    {MegaSecs, Secs, MicroSecs} = erlang:now(),
+    {MegaSecs, Secs, MicroSecs} = os:timestamp(),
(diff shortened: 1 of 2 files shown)


src/epgsql_wire.erl
```diff
@@ -209,6 +209,6 @@ encode_list(L) ->
     <<(byte_size(Bin)):?int32, Bin/binary>>.
 
 encode_standby_status_update(ReceivedLSN, FlushedLSN, AppliedLSN) ->
-    {MegaSecs, Secs, MicroSecs} = erlang:now(),
```
seriyps: erlang:now() is deprecated http://erlang.org/doc/apps/erts/time_correction.html and it throws compilation warnings in 18 and 19 Erlang's
http://tryerl.seriyps.ru/#id=4df3

seriyps: Here for example https://travis-ci.org/epgsql/epgsql/jobs/176440819#L328

davidw: I figured as much.  I recall reading about some of the time changes - are you positive that that's the best call and not one of the monotonic ones?  It certainly works with all versions.

seriyps: monotonic versions are added on erlang 18, but epgsql supports versions from 15 to 19. So, this call is the only one supported on all VMs. We can try to use conditional compilation, but I don't think there is a reason for this extra complexity.

seriyps:
@davidw thanks!

PR #105, opened 2016-11-16
Fix squery_row, equery_row typespecs. Fixes #68

seriyps:
Minor typespec fix. It's like #68, but fixes both row types.


seriyps:
@davidw thanks again

PR #106, opened 2016-11-17
Fix typespecs, related to date and time. Fixes #60.

### LINKED ISSUE #60: Error in data conversion
Hello,

I've found unexpected behaviour in timestamp conversion:

```
2> epgsql:equery(C1, "select * from test").
{ok,[{column,<<"test">>,timestamp,8,-1,1}],
    [{{{2015,3,2},{19,14,1.4729999999999999}}}]}
3> epgsql:squery(C1, "select * from test").
{ok,[{column,<<"test">>,timestamp,8,-1,0}],
    [{<<"2015-03-02 19:14:01.473">>}]}
```

So I expect 1.473 first query, but see something different, possibly convert to float error.

seriyps:
Actualize date and time typespecs.
The problem is that erlang `calendar` and postgresql `timestamp` have a bit different data ranges, plus, epgsql returns time seconds as floats.

https://www.postgresql.org/docs/current/static/datatype-datetime.html

Fixes #60

### DIFF
--- a/src/epgsql.erl
+++ b/src/epgsql.erl
@@ -29,7 +29,8 @@
 -export_type([connection/0, connect_option/0,
               connect_error/0, query_error/0,
               sql_query/0, bind_param/0, typed_param/0,
-              squery_row/0, equery_row/0, reply/1]).
+              squery_row/0, equery_row/0, reply/1,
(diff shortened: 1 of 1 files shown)


seriyps:
So, since there is no opposition to this change, I think it could be merged.

seriyps:
Thanks!

PR #108, opened 2016-11-18
Make it possible to pass connect options as `map()`

seriyps:
Maps can be used only on VM versions, that supports them, so, conditional compilation used.

For now maps are just converted to proplists early. But if at some moment we decide to drop support for Erlang < 17, we may start use maps internally.

### DIFF
--- a/src/epgsql.erl
+++ b/src/epgsql.erl
@@ -24,9 +24,10 @@
          sync_on_error/2,
          standby_status_update/3,
          start_replication/5,
-         start_replication/6]).
+         start_replication/6,
+         to_proplist/1]).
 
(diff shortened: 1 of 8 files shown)


src/epgsql.erl
```diff
@@ -37,15 +38,39 @@
...
+          replication => string()}.
+-else.
+-type connect_opts() :: [connect_option()].
+-endif.
+
+-type connect_error() ::
+        #error{}
+      | {unsupported_auth_method, atom()}
+      | invalid_authorization_specification
+      | invalid_password.
```
seriyps: This isn't related to maps, but fixed just while noticed

PR #113, opened 2016-11-23
Decode more error extra fields. Fixes #110

### LINKED ISSUE #110: Return all error details
This TODO https://github.com/epgsql/epgsql/blob/3.2.0/src/epgsql_wire.erl#L59

Now epgsql decodes only `detail, hint, position`.

But there are much more available now:
https://www.postgresql.org/docs/current/static/protocol-error-fields.html

seriyps:
Just return more details about error

https://www.postgresql.org/docs/9.5/static/protocol-error-fields.html

Fixes #110


seriyps:
Thanks!

Issue #86, opened 2015-12-16
(Minor) Inconsistency in delete ... returning


aherranz:
I think there is an inconsistent behaviour in delete...returning statement when no rows are deleted:

``` erlang
16> pgapp:squery(pre2,"delete from email_lnk returning *").
{ok,2,
    [{column,<<"id">>,text,-1,-1,0},
     {column,<<"created">>,int8,8,-1,0},
     {column,<<"email">>,text,-1,-1,0},
     {column,<<"tries">>,int4,4,-1,0}],
    [{<<"_6pj">>,<<"1450282807">>,<<"aherranz@...">>,
      <<"0">>},
     {<<"MKPh">>,<<"1450282813">>,<<"aherranz@...">>,
      <<"0">>}]}
17> pgapp:squery(pre2,"delete from email_lnk returning *").
{ok,0}
```

Since I am using _returning_ I would expect the second expression to return

``` erlang
{ok,0,
    [{column,<<"id">>,text,-1,-1,0},
     {column,<<"created">>,int8,8,-1,0},
     {column,<<"email">>,text,-1,-1,0},
     {column,<<"tries">>,int4,4,-1,0}],
    []}
```

Am I wrong?

seriyps:
Probably epgsql returns exactly what postgresql server returns to it. Should ensure...

aherranz:
```sql
email-> delete from email_lnk returning *;
  id  |  created   |        email         | tries 
------+------------+----------------------+-------
 bkAH | 1450305670 | aherranz@.........   |     0
 1Poy | 1450305673 | aherranz@.........   |     0
 YmjV | 1450305675 | angel.herranz@...... |     0
(3 rows)

DELETE 3
email=> delete from email_lnk returning *;
 id | created | email | tries 
----+---------+-------+-------
(0 rows)

DELETE 0
email=> delete from email_lnk returning *;
 id | created | email | tries 
----+---------+-------+-------
(0 rows)

DELETE 0
email=> delete from email_lnk;
DELETE 0
```

seriyps:
Ok, I've just realized that your problem may be related to pgapp, not epgsql driver.
Or it was resolved eventualy some time ago.

I checked at current `devel` branch:

```
> {ok, C} = epgsql:connect(#{host => "localhost", username => "postgres", password => "postgres", port => 5433}).
> epgsql:equery(C, "CREATE TEMPORARY TABLE t(c integer)").
{ok,[],[]}
[...]

PR #115, opened 2016-11-30
Fix `UPDATE/DELETE ... RETURNING ..` returning 0 rows

seriyps:
It's fix for #86 

So, now `UPDATE ... RETURNING ..` and `DELETE ... RETURNING ..` (and `INSERT .. RETURNING ..`, but I can't imagine such situation) will always return 4-tuple `{ok, Count, Columns, Rows}` from `squery`/`equery`.

Before this fix it was returning `{ok, 0}` when 0 rows were affected and this behaviour was inconsistent.

### DIFF
--- a/src/epgsql_sock.erl
+++ b/src/epgsql_sock.erl
@@ -730,7 +730,8 @@ on_message({?COMMAND_COMPLETE, Bin}, State0) ->
     Command = command_tag(State),
     Notice = {complete, Complete},
     Rows = lists:reverse(State#state.rows),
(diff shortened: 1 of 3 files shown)


davidw:
Any doc changes needed?

seriyps:
@davidw no, this behavior wasn't mentioned anywhere

aherranz:
Agree, no doc needed to be changed, but the reason is that the documentation (comments on `ok_reply`definition)  was already consistent with this implementation ;)

PR #120, opened 2017-02-03
Add int8range type support.

ElectronicRU:
Additionally, turn int4range and int8range into first-class binary types.

### DIFF
--- a/test/epgsql_tests.erl
+++ b/test/epgsql_tests.erl
@@ -965,6 +965,17 @@ range_type_test(Module) ->
       end,
       []).
 
+range8_type_test(Module) ->
+    with_min_version(
+      Module,
+      9.2,
+      fun(_C) ->
+          check_type(Module, int8range, "int8range(10, 20)", <<"[10,20)">>,
+                     [{1, 58}, {-1, 12}, {-985521, 5412687}, {minus_infinity, 0},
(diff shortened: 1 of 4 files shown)


ElectronicRU:
For some reason replication tests fail on OTP 19. May be a transient error, not experienced enough to know.

test/epgsql_tests.erl
```diff
@@ -959,7 +959,18 @@ range_type_test(Module) ->
...
+                      {984655, plus_infinity}, {minus_infinity, plus_infinity}])
+      end,
+      []).
+
+range8_type_test(Module) ->
+    with_min_version(
+      Module,
+      9.2,
+      fun(_C) ->
+          check_type(Module, int8range, "int8range(10, 20)", {10, 20},
```
seriyps: It would be nice to test values greater than int32. Ideally - upper and lower bounds.

seriyps:
@MikhailKalashnikov do you have a idea why replication tests fail here?

seriyps:
Thanks!

seriyps:
It looks like we have some race-condition or timing issue in our replication tests. Will investigate later.

ElectronicRU:
@seriyps looks to me like it might be caused by two replication tests for different Erlang versions coinciding in time.

PR #121, opened 2017-02-04
Use `ssl_opts` when trying to make `ssl` connection

danilagamma:
Otherwise `connect()` fails with something like:
```
** exception exit: {ssl_negotiation_failed,{options,{password,<<"my-password">>}}}
```

@seriyps please review.

### DIFF
--- a/src/epgsql_sock.erl
+++ b/src/epgsql_sock.erl
@@ -381,7 +381,8 @@ start_ssl(S, Flag, Opts, State) ->
     {ok, <<Code>>} = gen_tcp:recv(S, 1, Timeout),
     case Code of
         $S  ->
-            case ssl:connect(S, Opts, Timeout) of
+            SslOpts = proplists:get_value(ssl_opts, Opts, []),
+            case ssl:connect(S, SslOpts, Timeout) of
                 {ok, S2}        -> State#state{mod = ssl, sock = S2};
                 {error, Reason} -> exit({ssl_negotiation_failed, Reason})
             end;


seriyps:
Looks good. I wonder why we have `ssl_opts` in README which is not used anywhere until now...

Please, wait while we investigate problem with failed unit tests.

seriyps:
@danilagamma it seems that travis isn't able to rebuild pull-request with new version of .travis.yml in a target branch https://stackoverflow.com/a/17624403/410556.
It's suggested that PR should be closed and reopened, so, I'll try to do that.

seriyps:
Ok, now it looks like build fails because of your changes but not because builds are broken. Do you have any idea why?
Probably we missed something during migration to common test and rebar3?

seriyps:
Ok, it seems that it was me who introduced this wrong typespec here 8cf30023029cbcf8d2dd0445bea0b8b3c6176fbe. But in reality ssl options like `{keyfile, File("epgsql.key")}, {certfile, File("epgsql.crt")}` are just mixed with other `epgsql:connect` options which isn't nice.

We definitely have to fix this.

@danilagamma may you please change this line https://github.com/epgsql/epgsql/blob/devel/test/epgsql_SUITE.erl#L240

like this:

```diff
-    [{ssl, true}, {keyfile, File("epgsql.key")}, {certfile, File("epgsql.crt")}]).
+    [{ssl, true}, {ssl_opts, [{keyfile, File("epgsql.key")}, {certfile, File("epgsql.crt")}]}]).
```

and push to this PR as a separate commit?

seriyps:
Ok, I updated it myself, your changes were merged. Thanks! :+1:

PR #125, opened 2017-05-02
Fix for Erlang/OTP 20.0

okeuday:
Avoid conflict with new floor function.

### DIFF
--- a/src/epgsql_fdatetime.erl
+++ b/src/epgsql_fdatetime.erl
@@ -107,7 +107,7 @@ now2f({MegaSecs, Secs, MicroSecs}) ->
 tmodulo(T, U) ->
     Q = case T < 0 of
         true  -> ceiling(T / U);
-        false -> floor(T / U)
+        false -> flooring(T / U)
     end,
     case Q of
         0 -> {T, Q};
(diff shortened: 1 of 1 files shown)


davidw:
Hi, could you do this one against the 'devel' branch?  We should probably then release soonish to be ready for 20.0.  It'd be kind of nice if it could optionally use the OTP version of floor, too.

okeuday:
Won't be able to get time to do that.

davidw:
@seriyps have a look?

seriyps:
@davidw looks good

davidw:
Still like to conditionally use the native floor for 20 and hacked up an ugly version of it.

```
-ifndef(OTP_RELEASE).
flooring(X) ->
    T = erlang:trunc(X),
    case (X - T) of
        N when N < 0 -> T - 1;
        N when N > 0 -> T;
        _            -> T
    end.
-else.
-ifdef(?OTP_RELEASE >= 20).
flooring(X) ->
    floor(X).
-else.
flooring(X) ->
    T = erlang:trunc(X),
    case (X - T) of
        N when N < 0 -> T - 1;
        N when N > 0 -> T;
        _            -> T
    end.
-endif.
-endif.
```

But... yuck.

seriyps:
@davidw also smth like `erlang:function_exported(erlang, crc32, 1)` but I'm not sure if it's more effective.

PR #129, opened 2017-06-19
Cancel query when timeout is reached

robertoaloi:
This is a draft for how queries could be canceled on a given timeout, something that in the current `epgsql` implementation can only be achieved via an explicit `SET statement_timeout = X`.
This change adds support for timeouts for squery, equery, execute and prepared_query.
The reason for the explicit names (i.e. `xxx_with_timeout`) is not to break backward compatibility with the existing API.
The reason for having the tests in CT format separated from the rest is that I could not get the test DB up and running (the scripts assume a specific version of Postgres installed) and I run the CT suite against a docker container, instead.
I have plans to contribute the docker container and to eventually reorganize the tests a bit better.


seriyps:
A few notes:

1. Do you think we should use gen_server's mechanism of timeouts? Probably we may implement it inside `epgsql_sock` gen_server by some custom timer and this way make it universal even for asynchronous APIs. Plus we'll get rid of this `try gen_server:call(...) catch ...`, which values because of:
2. Do you think it's safe to not flush possible gen_server's replies after all? I see there is a race-condition between `gen_server:call`'s timeout and time when `epgsql:cancel` returns. I mean by the time when `cancel` will stop postgresql process it may just return response
We may try to reproduce it by sending queries like `epgsql:execute_with_timeout("SELECT pg_sleep(0.5)", [500])` when timeout is very close to execution time. This may be even simplified if `epgsql_sock`'s mailbox is overflowed.
```
|-------|--|-----------|
^ epgsql:call
        ^gen_server's timeout
        ^send epgsql:cancel
           ^original query finishes and response sent to caller
                        ^epgsql:cancel finishes
```
3. Is it correct to "fake" error record? I.e. not to return what exactly was returned by postgresql server but construct it manualy
4. I don't remember... Is it should be possible to execute more queries through this connection after query was cancelled? Maybe we should test that connection is still usable after such timeout error by executing more queries through it?

seriyps:
Regarding 1st note you may look how timeouts are implemented in https://github.com/devinus/poolboy (plus https://github.com/devinus/poolboy/issues/21) and https://github.com/seth/pooler.
Poolboy uses gen_server's timeout while pooler implements own mechanism.
1st one is simpler and more precise, while pooler's is more reliable (no race conditions).

seriyps:
Regarding 2nd note:

```
+---------+                          +-------------+    +-------------+
| Caller  |                          | epgsql_sock |    | PostgreSQL  |
+---------+                          +-------------+    +-------------+
     |                                      |                  |
     | gen_server:call with timeout         |                  |
     |------------------------------------->|                  |
     |                                      |                  |
     |                                      | query            |
     |                                      |----------------->|
     |                                      |                  | ------------------\
     |                                      |                  |-| executing query |
     |                                      |                  | |-----------------|
     | ------------------------------\      |                  |
     |-| got gen_server:call timeout |      |                  |
     | |-----------------------------|      |                  |
     |                                      |                  | ---------------------\
     |                                      |                  |-| execution finished |
     |                                      |                  | |--------------------|
     |                                      |                  |
     | epgsl:cancel                         |                  |
     |------------------------------------->|                  |
     |                                      |                  |
     |                                      |          results |
     |                                      |<-----------------|
     |                                      |                  |
     |                                      | cancel           |
     |                                      |----------------->|
     |                                      |                  |
     |                        results (!!!) |                  |
     |<-------------------------------------|                  |
     |                                      |                  |
     |                                      |        cancel ok |
     |                                      |<-----------------|
     |                                      |                  |
     |                            cancel ok |                  |
     |<-------------------------------------|                  |
     |                                      |                  |
```

test/epgsql_SUITE.erl
```diff
@@ -0,0 +1,115 @@
...
+        , execute_timeout/1
+        , prepared_query_timeout/1
+        , squery_timeout/1
+        ]).
+
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+%% Includes
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+-include_lib("common_test/include/ct.hrl").
+-include_lib("stdlib/include/assert.hrl").
```
seriyps: It looks like `stdlib/include/assert.hrl` only available from Erlang 18, but our travis.yml includes even Erlang 15 currently.
AFAIK, it's ok to use `eunit/include/eunit.hrl` instead.

robertoaloi:
Thanks for the quick reply.

1. No, I don't. Consider this PR as a draft to simply start the discussion. I will check the `epgsl_sock` module to see how this could be implemented there. I think your comment was truncated, though.
2. Will look into that.
3. Probably not, but we cannot really pass the return of the `cancel`, since that's an `ok` message. We could use something like `{error, timeout}`, but that will break the existing contract, though, which is why I faked the timeout for now.
4. That's something that bother me, too. Need to investigate this.
Extra: will convert to good, old `eunit.hrl`.

seriyps:
I think notes 1 and 2 are almost about the same. If we choose a correct timeout implementation there will be no need in flushing. My note n1 wasn't truncated, it just was continued as note n2. Sorry for confusion.
I woud suggest you to look at pooler's implementation of timeout. I think it fits our needs well.
3. I guess there are no contract for `*with_timeout` functions cuttently because we don't have any of them in our APIs yet :-) also it looks like our current apis are underspecified. Error spec covers nat all cases. See for example `epgsql_sock:flush_queue/2` which may produce return smth like `{error, sock_closed}`
4. Yep. Some investigation needed. Maybe we already have smth in our tests?
Sorry, I'm in a train now and writing from my phone.

robertoaloi:
Second iteration pushed.

* The timer management has been pushed inside `epgsql_sock`
* The timer is started when the request is added to the queue, canceled when the request is removed
* The timeout response is not faked anymore, but real
* The new proposed API has been changed from `xxx_with_timeout` to `timed_xxx`
* A timed variant of the API has been added only to the synchronous API for now

Please note that I added a `command` and a `timed_command` API to `epgsql_sock`, so that `epgsql` does not have to use `gen_server:call` directly.

A few things I noticed looking at the `epgsql_sock` code.

* The `gen_server` basically implements a state machine, which makes the code a bit hairy. Any reason why a `gen_server` has been preferred to a `gen_fsm`?
* Interaction with clients and with socket could be split into multiple processes to simplify the code
* Having a global timer when we have multiple requests in the queue sounds fragile to me. On the other hand, the `backend` is global, too. It seems like `epgsql` assumes that each `epgsql_sock` server is only used by one single process...

src/epgsql_sock.erl
```diff
@@ -89,13 +91,20 @@
                 repl_feedback_required,
                 repl_cbmodule,
                 repl_cbstate,
-                repl_receiver}).
+                repl_receiver,
+                timer}).
```
seriyps: Yep, looks like single timer may be not enough. Asynchronous interfaces `epgsql{a,i}` allow request pipelining like
```
Ref1 = epgsqla:timed_equery(...),
Ref2 = epgsqla:timed_equery(...),
Ref3 = epgsqla:timed_equery(...),
Res1 = recv .. Ref1 ...,
Res2 = recv .. Ref2 ...,
Res3 = recv .. Ref3 ...,
```
Pipelining in PostgreSQL works the same way as in HTTP pipelining https://en.wikipedia.org/wiki/HTTP_pipelining

But we have a kind of dilemma there, because, AFAIK, if we have, eg, 3 pending pipelined queries and only 1st of them have timeout, if timeout on this query reached and we send `cancel`, all pending requests will generate `query_canceled`. I think this behaviour is acceptable, but should be documented.
If we have 3 pipelined requests and each of them with timeout, then we have to decide what to do: keep only 1st active timer or have 3 timers in the mode 'the one which fires earlier 1st wins'...

-------------------------------

I think for beginning we may just state that asynchronous interface with timeouts in pipelining mode have limited functionality and must be used with care and full understanding.

src/epgsql_sock.erl
```diff
@@ -140,25 +149,37 @@ handle_call({standby_status_update, FlushedLSN, AppliedLSN}, _From,
     send(State, ?COPY_DATA, epgsql_wire:encode_standby_status_update(ReceivedLSN, FlushedLSN, AppliedLSN)),
     {reply, ok, State#state{repl_last_flushed_lsn = FlushedLSN, repl_last_applied_lsn = AppliedLSN}};
 
+handle_call({timed, Command, Timeout}, From, State) ->
+  Timer = erlang:start_timer(Timeout, self(), timed_command),
```
seriyps: I think if we decide to have a single timer, it may be reasonable to check if there is already active timer and cancel it (probably with `error_logger` warning). Or check which one will fire earlier and choose it.

src/epgsql_sock.erl
```diff
@@ -140,25 +149,37 @@ handle_call({standby_status_update, FlushedLSN, AppliedLSN}, _From,
...
 handle_cast(stop, State) ->
     {stop, normal, flush_queue(State, {error, closed})};
 
-handle_cast(cancel, State = #state{backend = {Pid, Key},
-                                   sock = TimedOutSock}) ->
+handle_cast(cancel, State) ->
+  do_cancel(State),
+  {noreply, State}.
+
+do_cancel(State) ->
```
seriyps: Please, move `do_cancel` implementation somewhere below `terminate` callback. I think it's not ok to mix internal functions and gen_server's callbacks.

src/epgsql_sock.erl
seriyps: UPD: I made some investigation and found following things:

* `epgsql:cancel/1` cancels only ONE request from pipeline (which is currently running one)
* connection is still usable after cancel

```erlang
2> {ok, C} = epgsql:connect(#{host => "localhost", port => 5433, username => "wgm", password=>"wgm", database=>"wgm1"}).
{ok,<0.142.0>}
% 3 commands are put in pipeline
19> epgsqla:squery(C, <<"SELECT pg_sleep(100)">>).
#Ref<0.0.1.1136>
20> epgsqla:squery(C, <<"SELECT pg_sleep(100)">>).
#Ref<0.0.1.1141>
21> epgsqla:squery(C, <<"SELECT pg_sleep(100)">>).
#Ref<0.0.1.1146>
% cancel 1st 2 of them
22> epgsqla:cancel(C).                            
ok
23> epgsqla:cancel(C).
ok
% got 2 'query_cancelled' responses (please note references order: #Ref<0.0.1.1136> then #Ref<0.0.1.1141> then #Ref<0.0.1.1146>)
24> flush().                                      
Shell got {<0.142.0>,#Ref<0.0.1.1136>,
           {error,{error,error,<<"57014">>,query_canceled,
                         <<"canceling statement due to user request">>,
                         [{file,<<"postgres.c">>},
                          {line,<<"3003">>},
                          {routine,<<"ProcessInterrupts">>}]}}}
Shell got {<0.142.0>,#Ref<0.0.1.1141>,
           {error,{error,error,<<"57014">>,query_canceled,
                         <<"canceling statement due to user request">>,
                         [{file,<<"postgres.c">>},
                          {line,<<"3003">>},
                          {routine,<<"ProcessInterrupts">>}]}}}
ok
% cancel 3rd request
25> epgsqla:cancel(C).
ok
% got 3rd 'query cancelled'
26> flush().          
Shell got {<0.142.0>,#Ref<0.0.1.1146>,
           {error,{error,error,<<"57014">>,query_canceled,
                         <<"canceling statement due to user request">>,
                         [{file,<<"postgres.c">>},
                          {line,<<"3003">>},
                          {routine,<<"ProcessInterrupts">>}]}}}
ok
% following cancels are noops
27> epgsqla:cancel(C).
ok
28> flush().          
ok
% connection is still usable
29> epgsql:squery(C, <<"SELECT pg_sleep(1)">>).   
{ok,[{column,<<"pg_sleep">>,void,4,-1,0}],[{<<>>}]}
```

It means that if we have 2 requests in a pipeline with different timeouts:

```
epgsqla:timed_squery(C, "SELECT pg_sleep(10)", 15000)
epgsqla:timed_squery(C, "SELECT pg_sleep(10)", 1000)
```
We expect 1st one to terminate successfully but the 2nd one with error. But we'll get absolutely different result: 1st query will be cancelled by 2nd timer.

I don't see any easy solution for this, so, we, probably, may just leave your curent implementation for now, but don't forget to document this behaviour of async interface. Pipelining is not so widely used feature, but it adds a bit of headache...

src/epgsql_sock.erl
```diff
@@ -89,13 +91,20 @@
...
 %% -- client interface --
 
 start_link() ->
     gen_server:start_link(?MODULE, [], []).
 
+command(C, Command) ->
+  gen_server:call(C, Command).
+
+timed_command(C, Command, Timeout) ->
+  gen_server:call(C, {timed, Command, Timeout}).
```
seriyps: It's definitely discussible, but we may imagine situation when `epgsql_sock`'s mailbox is overflowed and this `{timed ...` message stays in a mailbox for some significant amount of time. So, when it reaches `handle_call({timed...) -> erlang:start_timer(Timeout, ...)` we may have a lot of time already passed.

As a some kind of solution we may convert relative `Timeout` to absolute time before `gen_server:call` and then convert it back on arrival inside `handle_call`. But I'm not sure if it worth it.

src/epgsql_sock.erl
```diff
@@ -89,13 +91,20 @@
...
+                repl_receiver,
+                timer}).
 
 %% -- client interface --
 
 start_link() ->
     gen_server:start_link(?MODULE, [], []).
 
+command(C, Command) ->
+  gen_server:call(C, Command).
```
seriyps: Ah, and default `Timeout` for `gen_server:call` is 5s. We should pass `infinity` here.

src/epgsql_sock.erl
seriyps: The same: `gen_server:call(C, {timed, ..}, infinity)`

seriyps:
> The gen_server basically implements a state machine, which makes the code a bit hairy. Any reason why a gen_server has been preferred to a gen_fsm?

Initialy it was. See https://github.com/wg/epgsql/blob/master/src/pgsql_connection.erl. But AFAIK, it was changed to `gen_server` when asynchronous APIs were added. @mabrek may tell more.

> Interaction with clients and with socket could be split into multiple processes to simplify the code

Initialy it was. See https://github.com/wg/epgsql/blob/master/src/pgsql_sock.erl and https://github.com/wg/epgsql/blob/master/src/pgsql_connection.erl.
I don't know the exact reason. Guess it's to avoid extra messagge passing. @mabrek may tell more.

> Having a global timer when we have multiple requests in the queue sounds fragile to me. On the other hand, the backend is global, too. It seems like epgsql assumes that each epgsql_sock server is only used by one single process...

Already answered in review comments. The only valid case for multiple requests in the queue is asynchronous API (epgsqla, epgsqli) with pipelining. But your current implementation doesn't provide asynchronous timed implementations. I think it's ok for beginning.

PR #130, opened 2017-06-21
Rebar3

egobrain:
Here is backport of some features from my custom driver fork.
I think that they are really useful.
- All test are moved to common_tests (eunits are not suitable for the purpose for which they are currently used)
- Droped all custom scripts like start/stop db, because ct start clean postgres database manually on each run.
- Implicit .travis.yml (to use it in custom forks)
- Fixed dialyzer warnings on erlang > r19
- Droped support of R15B03 (master branch already doesn't build)


seriyps:
Bravo! :+1: 

Is it possible to tweak CT console output such way that one can understand which API interface used now (epgsql, epgslqi, epgsqla)?
For example here https://travis-ci.org/epgsql/epgsql/jobs/245562178#L328 we have (almost) each test case executed 3 times, but can't understand which interface used when.

egobrain:
@seriyps i've created [PR](https://github.com/ferd/cth_readable/pull/14) to cth_readable to fix this issue

seriyps:
@egobrain ah, ok. I just wonder if it's possible to do just on epgsql's side. But it looks like we can't control this. Ok, not a major problem.

I think it would be nice to mention in a README that it's possible to run tests on exact postgres version by changing $PATH like

```
PATH=$PATH:/usr/lib/postgresql/9.5/bin/ make test
```

Please, add this and if nobody else has any notes, I'm ready to merge.

egobrain:
@seriyps Fixed. Review please

seriyps:
Thanks! Great work!

PR #131, opened 2017-07-08
Correct specs to reflect tcp/socket errors and sock_closed

define-null:
I added a few cases that were not specified in the specs. Also changed one rare case to return `{inet_error, term()}` instead of just `term()`, so that those errors would be distinguishable. But actually instead of having those 3 different cases:
```
{sock_error, term()}
{inet_error, term()}
sock_closed
```
I would prefer smth that can be easily matched with `{error, {socket, _}}` for example. 
What do you think?

### DIFF
--- a/src/epgsql.erl
+++ b/src/epgsql.erl
@@ -29,7 +29,7 @@
          to_proplist/1]).
 
 -export_type([connection/0, connect_option/0, connect_opts/0,
-              connect_error/0, query_error/0,
+              connect_error/0, socket_error/0, query_error/0,
               sql_query/0, bind_param/0, typed_param/0,
               squery_row/0, equery_row/0, reply/1,
               pg_time/0, pg_date/0, pg_datetime/0, pg_interval/0]).
@@ -74,6 +74,9 @@
       | invalid_authorization_specification
       | invalid_password.
 -type query_error() :: #error{}.
+-type socket_error() :: {sock_error, term()}
+                      | {inet_error, term()}
+                      | sock_closed.
 
 %% Ranges are from https://www.postgresql.org/docs/current/static/datatype-datetime.html
 -type pg_date() ::
@@ -212,7 +215,8 @@ set_notice_receiver(C, PidOrName) ->
 get_cmd_status(C) ->
     epgsql_sock:get_cmd_status(C).
 
--spec squery(connection(), sql_query()) -> reply(squery_row()) | [reply(squery_row())].
+-spec squery(connection(), sql_query()) ->
+                reply(squery_row()) | [reply(squery_row())] | {error, socket_error()}.
 %% @doc runs simple `SqlQuery' via given `Connection'
 squery(Connection, SqlQuery) ->
     gen_server:call(Connection, {squery, SqlQuery}, infinity).
@@ -230,7 +234,8 @@ equery(C, Sql, Parameters) ->
             Error
     end.
 
--spec equery(connection(), string(), sql_query(), [bind_param()]) -> reply(equery_row()).
+-spec equery(connection(), string(), sql_query(), [bind_param()]) ->
+                reply(equery_row()) | {error, socket_error()}.
 equery(C, Name, Sql, Parameters) ->
     case parse(C, Name, Sql, []) of
         {ok, #statement{types = Types} = S} ->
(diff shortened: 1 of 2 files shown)


define-null:
@seriyps ping

seriyps:
BTW, I see that `inet_reply` isn't documented anywhere, so, probably, it's some legacy stuff (was introduced in 49d855570c46d861651c91a59efb45c5f212ac6f 5 years ago).
I think it's a dead code and may be removed safely. Will check.

seriyps:
Ok, I see that it's not a dead code; it's needed because we call inet_tcp port directly in `do_send` and it was done probably to avoid this `receive` https://github.com/erlang/otp/blob/OTP-20.0/erts/preloaded/src/prim_inet.erl#L430 (`gen_tcp:send() -> inet_tcp:send() -> prim_inet:send()`).

Since commit message doesn't contain any detailed explanation (probably was discowered by some hard profiling), I think this hack may be replaced with plain `Mod:send()` and this `handle_info({inet_reply..` clause may be removed.

@davidw @mabrek any thoughts?

src/epgsql.erl
```diff
@@ -259,8 +265,11 @@ parse(C, Sql) ->
 parse(C, Sql, Types) ->
     parse(C, "", Sql, Types).
 
--spec parse(connection(), iolist(), sql_query(), [epgsql_type()]) ->
-                   {ok, #statement{}} | {error, query_error()}.
+-spec parse(connection(), iolist(), sql_query(), [epgsql_type()]) -> Reply
+  when
+    Reply :: {ok, #statement{}}
+           | {error, query_error()}
+           | {error, socket_error()}.
```
seriyps: Maybe `{error, query_error() | socket_error()}`? I like both ways, but maybe you have some opinion on that?

src/epgsql.erl
```diff
@@ -269,8 +278,11 @@ parse(C, Name, Sql, Types) ->
 bind(C, Statement, Parameters) ->
     bind(C, Statement, "", Parameters).
 
--spec bind(connection(), #statement{}, string(), [bind_param()]) ->
-                  ok | {error, query_error()}.
+-spec bind(connection(), #statement{}, string(), [bind_param()]) -> Reply
+  when
+    Reply :: ok
+           | {error, query_error()}
+           | {error, socket_error()}.
```
seriyps: The same `{error, query_error() | socket_error()}`.

src/epgsql.erl
```diff
@@ -285,15 +297,19 @@ execute(C, S, N) ->
...
 
 -spec execute(connection(), #statement{}, string(), non_neg_integer()) -> Reply
-                                                                              when
+  when
       Reply :: {ok | partial, [equery_row()]}
              | {ok, non_neg_integer()}
              | {ok, non_neg_integer(), [equery_row()]}
-             | {error, query_error()}.
+             | {error, query_error()}
+             | {error, socket_error()}.
```
seriyps: The same `{error, query_error() | socket_error()}`

seriyps:
Regarding `{error, {socket, _Err}}` - what do you expect `_Err` will contain? If it will be `{error, {socket, {sock_error, term()} | {inet_error, term()} | sock_closed}}` it will be just too deep. Something like `{error, {socket, sock_error, term()} | {socket, inet_error, term()} | {socket, sock_closed, []} }` may work though.

seriyps:
Ok, @mabrek told me that this commit was introduced to be able to fix "slow gen_tcp:send()` by omitting selective receive in prim_inet (which is [still here](https://github.com/erlang/otp/blob/OTP-20.0/erts/preloaded/src/prim_inet.erl#L430)) on possible long message queue, and that this solution was borrowed from RabbitMQ https://github.com/rabbitmq/rabbitmq-common/blob/rabbitmq_v3_6_11/src/rabbit_writer.erl#L367-L389 (so, they also still using it).

More on that: https://github.com/erlang/otp/blob/OTP-20.0/erts/preloaded/src/prim_inet.erl#L430 and here http://erlang.org/pipermail/erlang-questions/2009-November/047523.html and here http://erlang.org/pipermail/erlang-questions/2014-March/078298.html
And there is a benchmark for this: https://gist.github.com/maxlapshin/2309864. After some modifications (removed `nosuspend`, send 10000 messages to `self()` before start and without that) it still shows significant improvement in performance ~20-50%.

As a conclusion, I think this hack may be left in place.

seriyps:
Looks like this was fixed long ago, so, closing

PR #114, opened 2016-11-24
Add get_cmd_status function

seriyps:
It's mostly for case, described in #112 and on this thread https://groups.google.com/forum/#!topic/epgsql/ePdg5-J6XZA

Also, `get_parameter` documented in README.

Not sure about the name, maybe somebody can suggest a better one?

Another open question is: should we return parsed status (as `atom() | {atom(), integer()}`) or raw value as binary string? Or maybe add option to read one of them (eg `get_cmd_status(C, raw), get_cmd_status(C, parsed)`).

PR #135, opened 2017-08-28
Return stack trace when rolling back a transaction

lukyanov:
What do you think about returning `erlang:get_stacktrace()` from `with_transaction` when the transaction is rolled back? Now it is hard to debug the code inside a transaction as `with_transaction` eats the stack trace.

### DIFF
--- a/src/epgsql.erl
+++ b/src/epgsql.erl
@@ -335,8 +335,7 @@ with_transaction(C, F) ->
     catch
         _:Why ->
             squery(C, "ROLLBACK"),
-            %% TODO hides error stacktrace
-            {rollback, Why}
+            {rollback, {Why, erlang:get_stacktrace()}}
     end.
 
 sync_on_error(C, Error = {error, _}) ->


seriyps:
We have a bunch of reports and PRs suggesting this:
https://github.com/epgsql/epgsql/issues/29
https://github.com/epgsql/epgsql/pull/111

I think that better option would be to add `with_transaction(Conn, Fun, Opts)` and add `Opts` will change the behavior of the function:

* Should it log exception
* Should it re-throw it or return? If return, should it return stacktrace also?
* Should it ensure that transaction was committed (see https://github.com/epgsql/epgsql/pull/114)
* Should it append options to BEGIN (eg `BEGIN [ WORK | TRANSACTION ] [ transaction_mode [, ...] ]`, see https://www.postgresql.org/docs/current/static/sql-begin.html)
* something else?

mcherep:
I created the pull request #111. I'll take a look at it again merging newer changes.

seriyps:
Fixed by #147

Issue #109, opened 2016-11-19
Implement pluggable datatypes


seriyps:
I've found, that Elixir postgresql driver uses pluggable datatype system.
https://github.com/elixir-ecto/postgrex/blob/master/lib/postgrex/types.ex
Sample plugin https://github.com/elixir-ecto/postgrex/blob/master/lib/postgrex/extensions/json.ex

I think it's a great idea.

- It allows users to implement customized encode/decode rules (eg, their preferred format of datetime, JSON, decimal etc)
- Allows users to easily handle their custom/private datatypes.
- New types may be added outside of epgsql main tree.
- epgsql code may became more 'beautiful' and generic :smiley: .

davidw:
Sounds good to me, although I would definitely include a bunch of them with the default distribution.

seriyps:
@ElectronicRU can you edit this one https://github.com/epgsql/epgsql/wiki/pluggable_datatypes ?

seriyps:
Things to consider:

* Type information for rows and parameters passed as `oid`: int32
* Mapping between OIDs and type information stored in `pg_type` pseudo-table. We have to have a local copy of this table to maintain OID<->Codec functions connection.
* Each(?) basic type OID have corresponding OID for array, containing this type
* Each 'CREATE TYPE' creates 2 new rows in `pg_type` (one for `my_type` itself, another for `my_type[]`)
* Previous point valid even for `enum`s. It would be good to be able to say 'decode all enums using this decoder`, as well as `decode this particular enum using this decoder`
* There are at least 3 composite types, which may contain any(?) datatype: `array`, `range` and `record`
* Date decoding algorithm depends on `integer_datetimes` parameter (see epgsql:get_parameter(C, <<"integer_datetimes">>)` and `epgsql_fdatetime.erl`, `epgsql_idatetime.erl` modules). So, codec should be aware of a value of this parameter. This parameter couldn't change during connection's lifetime according to https://www.postgresql.org/docs/current/static/protocol-flow.html#PROTOCOL-ASYNC
* Ranges. There are some built-in range types but range of any type may be created https://www.postgresql.org/docs/current/static/rangetypes.html. Each user-defined range type is added to `pg_type` and `pg_range` tables.
* Enums. Idea is the same as for ranges https://www.postgresql.org/docs/current/static/datatype-enum.html. Each user-defined enum type is added to `pg_type` and `pg_enum`
[...]

PR #138, opened 2017-09-10
Add command's packet protocol docs

seriyps:
I started to work on #109. This is a beginning of a serie of patches.

Here I just added not very formal description of sequences of PG protocol commands used to execute different epgsql commands.

### DIFF
--- a/src/epgsql_sock.erl
+++ b/src/epgsql_sock.erl
@@ -250,6 +250,12 @@ command({connect, Host, Username, Password, Opts}, State) ->
     end;
 
 command({squery, Sql}, State) ->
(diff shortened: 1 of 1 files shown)


davidw:
Nice work: one thought, though - would it make sense to keep these in an external doc?  Perhaps even in a format that could be used to generate a chart or something?  I'm more than happy to have the additional documentation as comments though, looks good!

seriyps:
Hey, while working on #109 I ended up with some significant refactoring; hope to submit it in next few days. This particular PR may become irrelevant then, so, let's put it into a hold for now!

seriyps:
Superseeded by #139

PR #139, opened 2017-09-13
Pluggable commands

seriyps:
Problems I'm trying to solve:

1. It's dificult to have command-specific state (eg, own state structure for 'equery' and another one for 'squery' and third one for 'connect'). This may be very useful to be able to keep "compiled" data row decoder while command executing. Another example is how we store `username` and `password` in connect.
2. Too many `if (Command == a or Command == b) do this elseif (...) do that` inside incoming message handling code (`on_message`).
3. Adding new commands is painfull because of (2)
4. If somebody needs a new command, he should implement it in a main repo, but it's not possible to have a command as a 'plugin'
5. Minor performance improvement may come from the fact that we don't put commands into a queue if there is only one command (which is most common case for 'epgsql' interface). But no benchmarks were made

Solutions:

1. Each command now may have it's private state which is passed to all callbacks
2. Each command handles only packets it knows. Server message processing loop is almost generic now. The only downside is that commands now have some duplicated code.
3. Add a new command now is as easy as create new module and implement `epgsql_command` behaviour
4. Those commands may be kept outside of a main repo. One can just use `epgsql_sock:sync_command/3` or `epgsql_sock:async_command/4` to execute them (or we may add 'command/3' APIs to our interface modules).
5. Currently executing command and it's state are kept directly as a field of `#state{}` structure, plus we eliminating all those `if Command == a or Command == b` which is also a good thing.

And it should make it easier to proceed with fixing #109


seriyps:
I'm reopening this pull-request so tests will run with a newer travis.ci settings

src/epgsql_sock.erl
```diff
@@ -100,6 +87,18 @@ close(C) when is_pid(C) ->
     catch gen_server:cast(C, stop),
     ok.
 
+-spec sync_command(epgsql:conection(), epgsql_command:command(), any()) -> any().
+sync_command(C, Command, Args) ->
+    gen_server:call(C, {command, Command, Args}, 1000).
```
seriyps: FIXME: timeout should be `infinity`

seriyps:
epgsql from this branch successfully works in live at my side-project (up to 60 rps) for ~1 week with no issues (but it's usage is relatively simple there). Plus I asked some people from community to test it on their projects and there were no issues related to this branch (some were found related to `pluggable-types`, but it's another story and they were fixed), so, I think this may be merged if no one is against it.

seriyps:
@davidw yep, I have one example: https://github.com/seriyps/epgsql/blob/equery2_command/src/commands/epgsql_cmd_equery2.erl (it's a equery that doesn't require 2 gen_server call's).
Unfortunately, no docs so far, but I hope to add them a bit later before the next release (as well as for pluggable types and `with_transaction/3`). Just don't have enough time right now.

PR #140, opened 2017-09-13
Fix travis builds; make tests more portable between PG versions

seriyps:
This should (hopefuly) fix currently failing travis builds.

### DIFF
--- a/test/epgsql_cth.erl
+++ b/test/epgsql_cth.erl
@@ -13,7 +13,7 @@ init(_Id, State) ->
     Start = os:timestamp(),
     PgConfig = start_postgres(),
     ok = create_testdbs(PgConfig),
-    error_logger:info_msg("postgres started in ~p ms\n",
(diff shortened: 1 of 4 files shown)


test/epgsql_cth.erl
```diff
@@ -13,7 +13,7 @@ init(_Id, State) ->
     Start = os:timestamp(),
     PgConfig = start_postgres(),
     ok = create_testdbs(PgConfig),
-    error_logger:info_msg("postgres started in ~p ms\n",
+    ct:pal(info, "postgres started in ~p ms\n",
```
davidw: Just out of curiosity, what's this change achieve?

seriyps: This makes it possible to change output verbosity by passing `--verbosity 0..100` option. Also, while it's not so relevant for epgsql (since we don't use error logger), but it makes it possible to distinguish output from tests and output from tested code.

PR #143, opened 2017-10-10
Add test case for connect network error

seriyps:
Just assert that in case of `gen_tcp:connect` error we first reply with error and then immediately stop epgsql_sock process.

### DIFF
--- a/test/epgsql_cast.erl
+++ b/test/epgsql_cast.erl
@@ -20,28 +20,30 @@
 
 connect(Opts) ->
     Ref = epgsqla:connect(Opts),
-    await_connect(Ref).
+    await_connect(Ref, Opts).
 
 connect(Host, Opts) ->
     Ref = epgsqla:connect(Host, Opts),
-    await_connect(Ref).
+    await_connect(Ref, Opts).
 
 connect(Host, Username, Opts) ->
     Ref = epgsqla:connect(Host, Username, Opts),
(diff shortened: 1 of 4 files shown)


test/epgsql_cast.erl
```diff
@@ -40,9 +40,9 @@ await_connect(Ref) ->
         {C, Ref, connected} ->
             {ok, C};
         {_C, Ref, Error = {error, _}} ->
-            Error;
-        {'EXIT', _C, _Reason} ->
```
seriyps: This clause was removed because it looks like the order of reply message and 'EXIT' message isn't guaranteed

test/epgsql_cast.erl
```diff
@@ -40,9 +40,9 @@ await_connect(Ref) ->
...
-            {error, closed}
+            Error
+    after 5000 ->
```
davidw: Is there any reason to make this configurable?

seriyps: @davidw it's possible - there is a `timeout` connect option, we may use it for this. But it's just a test module, not a public API and this timeout means failed test usually. Do you think somebody may need to change that?

seriyps: @davidw ok, I'll just do that using mentioned `timeout` connect option

seriyps: Done

PR #149, opened 2017-11-03
Fix reset.sql. Fixes #144

### LINKED ISSUE #144: reset.sql has 2 flaws
Hi Kreatorrr,

* in reset.sql, you can't suppress users before databases as they own DB objects ;
that triggers an error and all users persist after running this script.

* User "epgsql_test_replication" isn't deleted.

One more thing : test conditions would be better placed on top of the readme as users (including I !) usually skim through and miss this part - in Debian, as Pg is chopped in many parts, you fail some tests because of missing extension "postgis" that isn't part of the main package.

My 2¢ (1.70¢€;-p)

seriyps:
Fixes #144

### DIFF
--- a/test/data/reset.sql
+++ b/test/data/reset.sql
@@ -1,7 +1,8 @@
+DROP DATABASE epgsql_test_db1;
+DROP DATABASE epgsql_test_db2;
+
 DROP USER epgsql_test;
 DROP USER epgsql_test_md5;
 DROP USER epgsql_test_cleartext;
 DROP USER epgsql_test_cert;
-
-DROP DATABASE epgsql_test_db1;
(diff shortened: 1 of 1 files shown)


UmanShahzad:
Can we remove it? Only test_schema.sql is used:

```bash
$ git grep 'reset'
pluggable_commands.md:  so, client and server states are out-of-sync and we need to reset them.
$ git grep 'test_schema'
test/epgsql_cth.erl:        [Psql, Opts, "template1 < ", filename:join(?TEST_DATA_DIR, "test_schema.sql")]
```

seriyps:
Well, makes sense then.

PR #153, opened 2017-12-31
Pluggable types

### LINKED ISSUE #109: Implement pluggable datatypes

seriyps:
Fixes #109 . Motivation mostly described in #109. Some more info in newly added `pluggable_types.md`

How it works:

Epgsql requires mapping between oid (integer) and type name (eg `int4`, `jsonb` `varchar` etc). For standard types this mapping is stable and, before this change, was just hardcoded in epgsql code. Code that handled encoding/decoding of binary representation of data types also was hardcoded almost exclusively in `epgslq_binary` module.

What was done in this PR:

* We now maintain dynamic table with `Oid <-> TypeNameAndInfo` mapping (`epgsql_oid_db`); this table is prepopulated with hardcoded standard types mappings (`epgsql_binary:default_codecs()` and `epgsql_binary:default_oids()` on connection start-up and it can be updated by adding `codecs` connect option or by calling `update_type_cache/2`. This table is represented as `dict:dict()` on erlang < 18 and as `map()` one erlang 18+, because of performance. Table is kept in each epgsql_sock process's state and this means that crash reports will become larger and memory consumption may also grow.
* Each supported type now have corresponding module implementing `epgsql_codec` behaviour that knows how to encode/decode this type (actually, some modules know how to handle more than one type, eg one module for `int2`, `int4`, `int8`)
* For each equery/bind/prepared_query/execute_batch parameter we do a lookup over `epgsql_oid_db`s `Oid -> Codec` mapping and encode this parameter value as binary using codec's `encode/3` callback immediately.
* If query produces output (returns rows) we first "compile" decoder (see `epgsql_wire:build_decoder/2`): for each column we do lookup by `Oid -> Codec` table once, put codecs in a list that is parallel to a list of response's columns. Then for each row of result we are doing parallel iteration by lists of codecs and row's columns applying decoders to cells (see `epgsql_wire:decode_data/2`). This should give us significant speedup on result sets with many rows (we don't need to do `OID->Codec` table lookup for each row)

Backward-incompatible changes:

* `update_type_cache/2` signature changed
* typespecs were moved from `epgsql_geometry.hrl` to `ewkb.erl`
* `epgsql_type()` typespec was moved from `epgsql.hrl` to `epgsql.erl`
* Some of exported epgsql typespecs were deprecated
* we had a lot of types that we said we supported but in fact we were just bypassing binary data untouched (like `regproc`, `oidvector`, `pg_type_reltype`, `cash`, `anyelement` and a lot more). Now those types will be reported as unknown (`#column.type` will be `{unknown_oid, epgsql_oid_db:oid()}`). I don't think many users were relying on this.

Type encoding/decoding changes:

* Support for macaddr, macaddr8 was added
* Support for `nan` `plus_infinity` `minus_infinity` added to float

Tip: you may speedup connection setup by adding `{codecs, []}` connect option if you don't use hstore and PostGIS.


rebar.config
```diff
@@ -1,4 +1,5 @@
-{erl_opts, [{platform_define, "^[0-9]+", have_maps}]}.
+{erl_opts, [{platform_define, "^[0-9]+", have_maps},
+            {platform_define, "^(1[89])|^([2-9][0-9])", 'FAST_MAPS'}]}. % Erlang >=18
```
davidw: that's a hairy regexp!

seriyps:
Thanks, @davidw !

choptastic:
I know this has been closed for a while, but I just wanted to pop in here and say thanks for implementing this pluggable codec system. My little [sql_bridge](https://github.com/choptastic/sql_bridge) project has been depending on my own super-behind custom fork of epgsql, and this functionality allowed me to fully migrate the system to mainline epgsql without having to have my little hacks integrated into mainline epgsql. Kudos!

PR #154, opened 2018-01-04
Fix epgsqla and epgsqli API specs

zmstone:


seriyps:
Hi. We usualy do PRs over `devel` branch. devel branch merged it to master from time to time and then new tag / version added. But since your PR is small enough I think we can accept it like it is.

seriyps:
Would you mind to fix all the specs in epgsql, epgsqla, epgsqli where SQL query is an argument?

seriyps:
So yeah, can you, please, do the same with `epgsqli` module?

zmstone:
done.

seriyps:
Thanks!

Issue #142, opened 2017-10-06
Add support for PG10 scram-sha-256 auth method


seriyps:
Recently [released PostgreSQL 10](https://www.postgresql.org/docs/10/static/release-10.html) introduces new auth method scram-sha-256. I think it would be nice to have it in epgsql.

[Most detailed description](https://www.postgresql.org/docs/current/static/sasl-authentication.html)

[Packet format](https://www.postgresql.org/docs/current/static/protocol-message-formats.html)
We are interested in:
AuthenticationSASL
AuthenticationSASLContinue
AuthenticationSASLFinal
SASLInitialResponse
SASLResponse

[Packet flow](https://www.postgresql.org/docs/current/static/protocol-flow.html#idm46046840158032)

RFCs:
[rfc7677: SCRAM-SHA-256](https://tools.ietf.org/html/rfc7677)
[rfc5802: SCRAM SASL](https://tools.ietf.org/html/rfc5802)

seriyps:
Fixed by #156

Neustradamus:
@seriyps: You have done a good job in 2018 :)

Can you add the SCRAM-SHA-256-PLUS too?

-------------------------

SCRAM-SHA-256 has been added in PostgreSQL 10
SCRAM-SHA-256-PLUS variant (with TLS Binding) has been added in PostgreSQL 13
SCRAM-SHA-256 is selected by default in PostgreSQL 14

History:
- 10: https://www.postgresql.org/docs/10/auth-methods.html + https://www.postgresql.org/about/news/postgresql-10-released-1786/
- 11: https://www.postgresql.org/docs/11/auth-password.html
- 12: https://www.postgresql.org/docs/12/auth-password.html
- 13: https://www.postgresql.org/docs/13/auth-password.html + https://www.postgresql.org/about/news/postgresql-13-released-2077/
- 14: https://www.postgresql.org/docs/14/auth-password.html + https://www.postgresql.org/about/news/postgresql-14-released-2318/ 

On Microsoft.com:
- https://techcommunity.microsoft.com/t5/azure-database-for-postgresql/how-to-securely-authenticate-with-scram-in-postgres-13/ba-p/1548319

Linked to:
- https://github.com/scram-xmpp/info/issues/1

PR #156, opened 2018-02-15
Scram sha auth

seriyps:
See #142 for details.

https://github.com/sfackler/rust-postgres/blob/master/postgres-protocol/src/authentication/sasl.rs and https://github.com/npgsql/npgsql/pull/1769/files were used as references.

Code that implements authentication was reworked so it should be easier to add any new auth methods.

scram-sha-256 auth method was added to Postgresql 10 and not available on earlier postgresql releases.
However, it seems that travis-ci doesn't support PG10 yet https://github.com/travis-ci/travis-ci/issues/8537, so, we can't test this on travis, but I checked it locally.

### DIFF
--- a/include/protocol.hrl
+++ b/include/protocol.hrl
@@ -16,6 +16,7 @@
 -define(SIMPLEQUERY, $Q).
 -define(AUTHENTICATION_REQUEST, $R).
 -define(SYNC, $S).
+-define(SASL_ANY_RESPONSE, $p).
 
 %% Parameters
 
--- a/src/commands/epgsql_cmd_connect.erl
+++ b/src/commands/epgsql_cmd_connect.erl
@@ -12,18 +12,22 @@
 -type connect_error() ::
         invalid_authorization_specification
       | invalid_password
-      | {unsupported_auth_method, integer()}
+      | {unsupported_auth_method,
+         kerberosV5 | crypt | scm | gss | sspi | {unknown, integer()} | {sasl, [binary()]}}
+      | {sasl_server_final, any()}
       | epgsql:query_error().
 
 -include("epgsql.hrl").
 -include("protocol.hrl").
 
+-type auth_fun() :: fun((init | binary(), _, _) -> {send, byte(), iodata(), any()} | ok | {error, any()}).
+
 -record(connect,
         {opts :: list(),
-         auth_fun :: fun() | undefined,
+         auth_fun :: auth_fun() | undefined,
          auth_state :: any() | undefined,
          auth_send :: {integer(), iodata()} | undefined,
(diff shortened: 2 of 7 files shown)


seriyps:
@davidw thanks! Yep, me too. I just implemented it from specification with simplification for `normalize` function (because full implementation of that is huge). I may send a mail to our mail list to call for some more reviewers.

mmzeeman:
It looks ok, with one remark. 

Implementations of crypto protocols often go sideways with presumably simple things like generating and validating nonce's. 

People often just generate a random number, in this case an 80 bit random number. This means, because of the birthday paradox, that you have to watch 2**40 login sessions to actually see a nonce used twice!

Nonce's don't have to be random, they have to be unique. The combination of the nonce, with (in this case) the password should be used once and only once (in the entire system).  

Maybe 2**40 is probably enough uniqueness here, but maybe there is a simple way to improve this by mixing in a timestamp of some sorts.

Sometimes validating the uniqueness of incoming nonce's is also important. I really don't know if that is the case for this protocol.

seriyps:
Thanks, @mmzeeman ! That was very educational comment.

I added unique part to the nonce (it uses different mechanisms for erlang before 18 and >=18) and set random part size to 16 bytes.

I don't think we should add validation of uniqueness of the server's nonce... It might be too complex.

seriyps:
@davidw do you think this can be merged now?

davidw:
Looks good!

seriyps:
Thanks everybody

Neustradamus:
@seriyps: You have done a good job in 2018 :)

Can you add the SCRAM-SHA-256-PLUS too?

-------------------------

SCRAM-SHA-256 has been added in PostgreSQL 10
SCRAM-SHA-256-PLUS variant (with TLS Binding) has been added in PostgreSQL 13
SCRAM-SHA-256 is selected by default in PostgreSQL 14

History:
- 10: https://www.postgresql.org/docs/10/auth-methods.html + https://www.postgresql.org/about/news/postgresql-10-released-1786/
- 11: https://www.postgresql.org/docs/11/auth-password.html
- 12: https://www.postgresql.org/docs/12/auth-password.html
- 13: https://www.postgresql.org/docs/13/auth-password.html + https://www.postgresql.org/about/news/postgresql-13-released-2077/
- 14: https://www.postgresql.org/docs/14/auth-password.html + https://www.postgresql.org/about/news/postgresql-14-released-2318/ 

On Microsoft.com:
- https://techcommunity.microsoft.com/t5/azure-database-for-postgresql/how-to-securely-authenticate-with-scram-in-postgres-13/ba-p/1548319

PR #157, opened 2018-02-23
Re-generate epgsql_errcodes.erl

seriyps:
Just updating to latest version


generate_errcodes_src.sh
```diff
@@ -3,7 +3,7 @@
 # Used to generate epgsql_errcodes.erl
 #
 ERRFILE="https://raw.github.com/postgres/postgres/master/src/backend/utils/errcodes.txt"
```
davidw: Should have noticed that earlier, but should we tie it to a specific version or tag or something?

seriyps: I don't think that nakes much sense: it seems that this table is backwards-compatible. So they only add to it but not remove. It should be ok to sync with latest master from time to time

davidw: Ok, makes sense. Thanks!

PR #163, opened 2018-06-23
Compatibility with Erlang 21

seriyps:
erlang:get_stacktrace() was deprecated


seriyps:
I think we can release 4.1.0 after merging this one and #161 and then drop R16 support.
What do you think @davidw ?

Issue #162, opened 2018-06-23
Stop supporting Erlang R16


seriyps:
- Erlang R16B3 was released in 2014 and there were already 5 major Erlang releases since then.
- It's the last one currently supported Erlang release that doesn't have maps.
- We already have some "ifdef"s and hacks in rebar.config because of R16 (eg, elvis config).
- Latest rebar 3.6.0 doesn't support it.

seriyps:
Closed by #165

PR #164, opened 2018-06-23
Use R16-compatible rebar3 on R16

seriyps:
This is a small workaround for rebar3 R16 incompatibilities introduced after Rebar3 3.5.2.
But if we would decide to drop R16 support (see #162), this PR won't be needed


seriyps:
I think we can safely merge this one to make our `devel` branch green and if we later decide to drop R16 support this can be safely reverted or just cleaned-up.

PR #165, opened 2018-06-27
Drop R16 support; use maps as internal options datastructure

### LINKED ISSUE #162: Stop supporting Erlang R16

seriyps:
Fixes #162.
Some discussion was also there https://groups.google.com/forum/#!topic/epgsql/CVHt4Z0IX8U

* Drop R16 support and remove all the workarounds introduced to be able to work on R16
* Use maps as internal datastructure for options instead of proplists
* README was cleaned-up a little to reflect that maps are the preferred way to pass options to "connect"


davidw:
I'm a bit behind with the wider world of Erlang - is switching to maps happening in other projects too?

PR #168, opened 2018-08-23
Bugfix decoding of nested records and arrays

egobrain:
Currently in 4+ decoding of nested structures is broken:

```erlang repl
1> epgsql:equery(C, "select (1, (select (2,3)));", []).
{ok,[{column,<<"row">>,{unknown_oid,2249},2249,-1,-1,1}],
    [{{1,
       <<0,0,0,2,0,0,0,23,0,0,0,4,0,0,0,2,0,0,0,23,0,0,...>>}}]}
```

```erlang repl
2> epgsql:equery(C, "select (1, '{2,3}'::int[]);", []).

13:20:28.949 [error] CRASH REPORT Process <0.17684.0> with 0 neighbours crashed with reason: no function clause matching epgsql_codec_integer:decode(<<0,0,0,1,0,0,0,0,0,0,0,23,0,0,0,2,0,0,0,1,0,0,0,4,0,0,0,2,0,0,0,4,0,0,0,3>>, int4, []) line 36
```

### DIFF
--- a/test/epgsql_SUITE.erl
+++ b/test/epgsql_SUITE.erl
@@ -67,6 +67,7 @@ groups() ->
             hstore_type,
             net_type,
             array_type,
+            record_type,
             range_type,
             range8_type,
             custom_types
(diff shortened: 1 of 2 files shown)


seriyps:
Thanks! Can you add some description to PR? I think I got the idea, but not 100% sure.
Is that true, that before this PR it was able to decode only "scalar" types inside of records, because it was using too low-level functions to decode record's items?

test/epgsql_SUITE.erl
```diff
@@ -847,6 +848,17 @@ array_type(Config) ->
...
 
+record_type(Config) ->
+    Module = ?config(module, Config),
+    epgsql_ct:with_connection(Config, fun(C) ->
+        {ok, _, [{{1, 2}}]} = Module:equery(C, "select (1,2)", []),
+        {ok, _, [{{1, {2, 3}}}]} = Module:equery(C, "select (1, (select (2,3)))", []),
+        {ok, _, [{{1, [2, 3]}}]} = Module:equery(C, "select (1, '{2,3}'::int[])", []),
+        {ok, _, [
+            {{<<"test">>,[{1,<<"one">>},{2,<<"two">>}]}}
+        ]} = Module:equery(C, "select ('test', ARRAY(select (id, value) from test_table1))", [])
```
seriyps: nit: would be cool to have some comments describing what are we asserting in each case. Like "record inside record", "array nested in record" etc.

test/epgsql_SUITE.erl
```diff
@@ -847,6 +848,17 @@ array_type(Config) ->
         Select(jsonb, [<<"{}">>, <<"[]">>, <<"1">>, <<"1.0">>, <<"true">>, <<"\"string\"">>, <<"{\"key\": []}">>])
     end).
 
+record_type(Config) ->
+    Module = ?config(module, Config),
+    epgsql_ct:with_connection(Config, fun(C) ->
+        {ok, _, [{{1, 2}}]} = Module:equery(C, "select (1,2)", []),
+        {ok, _, [{{1, {2, 3}}}]} = Module:equery(C, "select (1, (select (2,3)))", []),
+        {ok, _, [{{1, [2, 3]}}]} = Module:equery(C, "select (1, '{2,3}'::int[])", []),
```
seriyps: May you also, please, try to use `?assertMatch` instead of just pattern-matching?

test/epgsql_SUITE.erl
egobrain: Fixed. Review please

egobrain:
> Is that true, that before this PR it was able to decode only "scalar" types inside of records, because it was using too low-level functions to decode record's items?

Yes. But in 3+ nested record works well.

seriyps:
Thanks!

PR #174, opened 2018-10-11
Simplify `platform_define` rebar configs

seriyps:
The main reason for this change is to make it possible to build epgsql on all supported OTP releases even without rebar and `paltform_define` applied.
But as a bonus, `platform_define` is now looks much simpler.


zmstone:
❤️

seriyps:
Thanks for reviewing this!

Issue #175, opened 2018-10-17
Crash when using epgsql:connect/1 with proplist as argument


AndreyKlimchuk:
Bug introduced in latest master (4.2.0)

```
{badmatch,
                          {error,
                           {{function_clause,
                             [{epgsql_cmd_connect,init,
                               [[{size,10},
                                 {host,"..."},
                                 {database,"..."},
                                 {username,"..."},
                                 {password,[]},
                                 {port,5432}]],
                               [{file,"src/commands/epgsql_cmd_connect.erl"},
                                {line,45}]},
                              {epgsql_sock,command_new,4,
                               [{file,"src/epgsql_sock.erl"},{line,256}]},
                              {gen_server,try_handle_call,4,
                               [{file,"gen_server.erl"},{line,661}]},
                              {gen_server,handle_msg,6,
                               [{file,"gen_server.erl"},{line,690}]},
                              {proc_lib,init_p_do_apply,3,
                               [{file,"proc_lib.erl"},{line,249}]}]}
```
Noticed when using pgapp

seriyps:
Hi. Good catch! Will fix

seriyps:
Fixed. Published as 4.2.1

AndreyKlimchuk:
Cool, thanks !

PR #176, opened 2018-10-17
Fix connect proplist

seriyps:
See #175 for detail. Bug was introduced in 4.2.0


davidw:
Maybe add a test to ensure things continue to work as advertised in the future?

seriyps:
@davidw good idea! Added and found bug (fixed)

seriyps:
Thanks! I will make 4.2.1 tag

PR #181, opened 2019-01-09
Add options to epgsql:start_replication

MikhailKalashnikov:
Add options to epgsql:start_replication. 
If decoding plugin filter some WAL records (for example plugin pgoutput and PG publications with some tables) align_lsn = True must be set, otherwise it will not be possible to stop PostgreSql DB correctly when logical replication is running.


seriyps:
would it be possible to add some tests that will showcase the mentioned problem?
Otherwise, looks good

MikhailKalashnikov:
I can't find a way to make an auto test for this case, because you actually do not see any error but just can't stop Postgres DB.
I've created an example https://github.com/MikhailKalashnikov/repl_epgsql_test 
which can be used to reproduce this case and see that it can be solved by setting a new option which I introduced in this pull request

seriyps:
I tried to read through postgresql logical replication docs for a while and still not feel 100% confident on how does it work in this case.
So, what are you saying is happening there is:

1. Replication plugin analyzes replication messages generated by postgresql server and may decide to not send some segments, so, some of `?COPY_DATA, <<?X_LOG_DATA, ...` messages will just not be sent to epgsql
2. But postgresql still expects from epgsql that it will report those missing segments as received

Thing that I'm not sure I understand is that docs say that `LSN` in `?COPY_DATA, <<?PRIMARY_KEEPALIVE_MESSAGE:8, LSN:?int64` message is "The current end of WAL on the server.", which might be far away forward from what we received in epgsql if we have some replication delay. Does that mean that we will skip all the not yet received wal segments if we will get "Primary keepalive message".

MikhailKalashnikov:
We will reset LSN only when receive "Primary keepalive message" with the special flag (which means that the client should reply to this message as soon as possible).
If epgsql has some replication delay it means that it keep receiving WALs in XLogData message. epgsql sends update to server (via Standby status update) after XLogData processing with last processed LSN. So PG server will receive updates and will not send "Primary keepalive message" with this special flag.
But in case of PG server shutdown there is a risk to skip not yet received wal segments.

However if you use epgsql replication to implement 'Consistent cache' (described in streaming.md) and re-create replication slots (or use temporary slots) and load all data from tables on an application startup then you do not have any risk to lose data.
Since epgsql may be used for different cases I implemented this behavior as an optional, so users can decide and set this align_lsn option accordingly.

MikhailKalashnikov:
I updated streaming.md and add detailed description of this new option

seriyps:
Ok, I'm going to merge this on monday if there will be no other objections.

seriyps:
Merged. Thanks!

PR #186, opened 2019-02-06
Handle errors in the batch command.

kape1395:
If an error occurs while executing a batch
of statements, the error is returned by PG and
all the remaining statements are not executed.
We need to handle that and return appropriate
error message in order to have result list to
match with the statements in the batch.

Before this change, the epgsql user received
an error on unexpected message 90 ("Z") from
the postgres server.
""

### DIFF
--- a/src/commands/epgsql_cmd_batch.erl
+++ b/src/commands/epgsql_cmd_batch.erl
@@ -70,10 +70,18 @@ handle_message(?COMMAND_COMPLETE, Bin, Sock,
                      {ok, Rows}
              end,
     {add_result, Result, {complete, Complete}, Sock, State#batch{batch = Batch}};
-handle_message(?READY_FOR_QUERY, _Status, Sock, #batch{batch = B} = _State) when
-      length(B) =< 1 ->
+handle_message(?READY_FOR_QUERY, _Status, Sock, #batch{batch = Batch} = _State) ->
     Results = epgsql_sock:get_results(Sock),
-    {finish, Results, done, Sock};
+    case Batch of
+        [] ->
+            {finish, Results, done, Sock};
+        [_|_] ->
+            % In the case of error, the remaining statements are not executed.
+            % We return {error, skipped} for them to have result list matching
+            % with the statement list in the batch.
+            Skipped = lists:duplicate(length(Batch), {error, skipped}),
+            {finish, Results ++ Skipped, done, Sock}
+    end;
 handle_message(?ERROR, Error, Sock, #batch{batch = [_ | Batch]} = State) ->
     Result = {error, Error},
     {add_result, Result, Result, Sock, State#batch{batch = Batch}};
--- a/test/epgsql_SUITE.erl
+++ b/test/epgsql_SUITE.erl
@@ -435,7 +435,7 @@ batch_error(Config) ->
     Module = ?config(module, Config),
     epgsql_ct:with_rollback(Config, fun(C) ->
         {ok, S} = Module:parse(C, "insert into test_table1(id, value) values($1, $2)"),
-        [{ok, 1}, {error, _}] =
+        [{ok, 1}, {error, _}, {error, skipped}] =
             Module:execute_batch(
               C,
               [{S, [3, "batch_error 3"]},


seriyps:
Hi,
Sorry for delay.
Can you, please, fix failing tests?

kape1395:
I would like to, but I'm not sure, where to look for a problem.
The root cause looks unrelated with my changes:

    {'EXIT',{ssl_negotiation_failed,{tls_alert,"certificate expired"}}}

I see the same errors on the devel branch.

seriyps:
True. Seems SSL certificates that are used in tests just expired. I will try to fix this and will be back

seriyps:
I fixed ssl cerificate issue. Can you, please, rebase your pull-request over current `devel` branch and force-push?

kape1395:
I have rebased my branch on devel.

The tests are still failing at the devel branch. First error on my machine is

````
info 2019-02-19 09:28:43.959
postgres: 2019-02-19 07:28:43.958 GMT [7357] ERROR:  could not open extension control file "/usr/share/postgresql/10/extension/postgis.control": No such file or directory
2019-02-19 07:28:43.958 GMT [7357] STATEMENT:  CREATE EXTENSION postgis;
````

then

````
info 2019-02-19 09:28:43.961
postgres: 2019-02-19 07:28:43.961 GMT [7357] ERROR:  type "geometry" does not exist at character 388
2019-02-19 07:28:43.961 GMT [7357] STATEMENT:  CREATE TABLE test_table2 (
          c_bool bool,
          c_char char,
........
----------------------------------------------------
info 2019-02-19 09:28:43.962
postgres: 2019-02-19 07:28:43.962 GMT [7357] ERROR:  relation "test_table2" does not exist
2019-02-19 07:28:43.962 GMT [7357] STATEMENT:  GRANT ALL ON TABLE test_table2 TO epgsql_test;
````

**Comment by seriyps** [comment]
On your locsl machine you should install postgis extension

вт, 19 февр. 2019 г., 8:32 Karolis Petrauskas notifications@github.com:

**Comment by kape1395** [comment]
OK, that helped. Now I see two errors related to batch execution. Will try to fix them.

**Comment by seriyps** [comment]
It's just one test. It's executed 3 times because we test each case over 3 API modules (epgsql, epgsqla, epgsqli). The error is there:

https://github.com/epgsql/epgsql/blob/20ece6fbcb6a84aaf04d8df108f96f264e675e7d/test/epgsql_SUITE.erl#L438

**Comment by kape1395** [comment]
I see. Should it return two or three entries in the list?
My intention was to return  `[{ok, 1}, {error, _}, {error, _}]` to have result list zippable with the request list.
Another approach would be to returns results until the first error.

**Comment by seriyps** [comment]
> Another approach would be to returns results until the first error.

Isn't it the current behaviour?

Yep, I think we can do `[{ok, 1}, {error, _}, {error, skipped}]`

**Comment by kape1395** [comment]
I pushed the fixed test, but it fails for the incremental transport.
The last result is lost here, https://github.com/epgsql/epgsql/blob/20ece6fbcb6a84aaf04d8df108f96f264e675e7d/src/epgsql_sock.erl#L408
I had no time to find proper solution for that yet.

**Comment by seriyps** [comment]
yep, I believe it might be bit trickier, because incremental protocol is more "verbose". It fails because this clause wasn't executed, but "result" is created

```
handle_message(?COMMAND_COMPLETE, Bin, Sock,
               #batch{batch = [{#statement{columns = Columns}, _} | Batch]} = State) ->
    Complete = epgsql_wire:decode_complete(Bin),
    Rows = epgsql_sock:get_rows(Sock),
    Result = case Complete of
                 {_, Count} when Columns == [] ->
                     {ok, Count};
                 {_, Count} ->
                     {ok, Count, Rows};
                 _ ->
                     {ok, Rows}
             end,
    {add_result, Result, {complete, Complete}, Sock, State#batch{batch = Batch}};
```
incremental protocol needs this `add_result` to work properly. I'm not sure if it's possible to emulate it with current API for `epgsql_command` behaviour. Probably we will need to extend it

egobrain:
Any news here ?
I've got same issue

seriyps:
Unfortunately I don't have time to work on this issue. Proposed solution do work for most of the cases except "incremental" API, so, I can propose to use a fork for now, if it's critical for you?

seriyps:
Alternative solution #228 for this problem was merged

PR #187, opened 2019-02-17
Update test certificates

seriyps:
Current tls certificates expired 2019-02-03, so, I re-generated them. Without this fix our tests are broken.

Also, I added test for cases when client certificate is invalid and have made handling of tls handshake errors bit more graceful.


generate_test_certs.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
```
davidw: Does this require bash or would /bin/sh work too?

seriyps: Good point! I updated it to use `/bin/sh` instead.

seriyps:
Thank you for review, @davidw

PR #188, opened 2019-03-14
Fix typespec since password may be a string() or a fun()

onno-vos-dev:
Dialyzer is not so happy due to the wrong spec :)

### DIFF
--- a/src/epgsql.erl
+++ b/src/epgsql.erl
@@ -44,11 +44,12 @@
 
 -type sql_query() :: iodata().
 -type host() :: inet:ip_address() | inet:hostname().
+-type password() :: string() | iodata() | fun( () -> iodata() ).
 -type connection() :: pid().
 -type connect_option() ::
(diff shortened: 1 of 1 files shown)


src/epgsql.erl
```diff
@@ -44,11 +44,12 @@
 
 -type sql_query() :: iodata().
 -type host() :: inet:ip_address() | inet:hostname().
+-type password() :: string() | fun().
```
seriyps: Can you, please, define `fun()` as `fun( () -> iodata() )`? Also, password can be not only a string but binary or any iodata as well

onno-vos-kivra: Updated. Since `iodata()` is `iolist() | binary()` I did not add `binary()` explicitly.

seriyps: @onno-vos-kivra perfect, thanks!

seriyps:
Thank you! :+1:

PR #191, opened 2019-04-10
Add type spec for epgsql:equery/3

onno-vos-dev:
I recently encountered a bug in one of my projects which would have been caught by dialyzer if we had a spec for `epgsql:equery/3`. I suspect that `epgsql:equery/3` used to call `epgsql:equery/4` (which does have a spec) but since it no longer has, dialyzer doesn't know the spec anymore. 

It might make sense to add a `warn_missing_spec` in order to enforce specs on all exported functions. @seriyps how do you feel about that? In that case, I can work on a PR that introduces specs for all exported functions and avoid these small PRs :) The thing though is that many of the exported functions are only used internally across epgsql modules so it'd be more for internal reasons than for users of this library.

### DIFF
--- a/src/epgsql.erl
+++ b/src/epgsql.erl
@@ -233,6 +233,8 @@ equery(C, Sql) ->
     equery(C, Sql, []).
 
 %% TODO add fast_equery command that doesn't need parsed statement
+-spec equery(connection(), sql_query(), [bind_param()]) ->
+                    epgsql_cmd_equery:response().
 equery(C, Sql, Parameters) ->
     case parse(C, "", Sql, []) of
         {ok, #statement{types = Types} = S} ->


seriyps:
Yeah, I used to add specs only to lowest-level functions for situations like

```erlang
my_fun(A) ->
   my_fun(A, [] = _SomeDefault).
-spec my_fun(type1(), type2()) -> type3().
my_fun(A, B) ->
  <..>.
```

Just because dialyzer is usually smart enough to infer types for `my_fun/1` from typespec of `my_fun/2`. But yep, it probably sneaked when I changed some orders in which functions call each other.

Regarding `warn_missing_spec`: as you said, it really makes sense to have it for API modules (`epgsql`, `epgsqla` and `epgsqli`), but for all other ones... well, if someone else will do all the dirty work, then I'm ok =)

onno-vos-dev:
> Regarding `warn_missing_spec`: as you said, it really makes sense to have it for API modules (`epgsql`, `epgsqla` and `epgsqli`), but for all other ones... well, if someone else will do all the dirty work, then I'm ok =)

Only way to then enforce this would be to have some custom check that allows for warn_missing_spec for specific modules. I'll fix up a PR that adds type specs for the 3 API modules for starters. 

Surely someone must have had the same requirement to only have a `warn_missing_spec` for some super-duper-imporant-module? :roll_eyes: I'll do some investigation and try and find out.

seriyps:
@onno-vos-dev looks like one can just add

```erlang
-compile(warn_missing_spec).
```
to mentioned erlang modules (so, it's actualy not dialyzer option, but compiler option)

onno-vos-dev:
I'll fix, you fine with that as a seperate PR or do you want me to fix this one and the others in one go?

seriyps:
Well, both works for me, so, up to you. I just added @davidw as a reviewer and would prefer to wait for his opinion at least for one day in case he have any objections

onno-vos-dev:
No problem. Merge this then :) At least maybe someone else won't run in the same problem as I did. I'll follow up with the other specs over the Easter weekend :+1:

seriyps:
Thanks, @onno-vos-dev , @davidw !

PR #194, opened 2019-05-21
Add Erlang 22 to travis; replace 21.0 with 21.3

seriyps:
Just added freshly released Erlang to travis builds


seriyps:
Thanks! Seems we can remove Erlang 17 from the list of supported versions and related maps workarounds as well!

PR #195, opened 2019-05-21
Drop erl17 support

seriyps:
This PR contains 3 not strictly related changes:

- Dropped support for Erlang 17
- Use PostgreSQL 10 server in travis tests
- erlexec test dependency is fetched from hex.pm and upgraded

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,6 @@ otp_release:
   - 20.3
   - 19.3
   - 18.3
-  - 17.5
 script:
   - make elvis
   - make test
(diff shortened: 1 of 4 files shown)


davidw:
Without looking at things carefully, is it worth putting out and/or documenting the last supported version, for anyone stuck on an old Erlang version?

seriyps:
> Without looking at things carefully, is it worth putting out and/or documenting the last supported version, for anyone stuck on an old Erlang version?

I can do in release notes when we will tag new epgsql release

seriyps:
Thank you!

PR #196, opened 2019-06-20
Support timeout prepared_query/4 API

turtleDeng:


seriyps:
If you are interested, I would recommend to try to complete #129 story, because it was quite close to be merged, but then abandoned.
If not, then I would recommend to use one of asynchronous APIs (it's usually ok to combine asynchronous and synchronous APIs with single connection). Smth like:

```erlang
{ok, C} = epgsql:connect(...),
{ok, _} = epgsql:parse(C, "inc", "select $1+1", []),
Ref = epgsqla:prepared_query(C, Statement, [TypedParameters]),
receive
  {C, Ref, Res} -> Res
after Timeout ->
    handle_timeout()
end.
```

But be ready that you still can receive result of your query later

turtleDeng:
Thank you very much and thank you for your advice.

PR #197, opened 2019-06-26
Allow configuring a json module with options for encode/decode in the json codec

UmanShahzad:


src/datatypes/epgsql_codec_json.erl
```diff
@@ -17,18 +17,47 @@
...
-%% TODO: JSON encode/decode `fun Mod:Name/1` / `{Mod, Name}` as option.
-%% Shall not pass `fun(_) -> .. end`, because of hot code upgrade problems.
-init(_, _) -> [].
+%% JsonMod shall either be a module with encode/1 and decode/1, where:
+%%
+%% - encode/1 expects erlang-formatted JSON, e.g. a map or a proplist,
+%%   and returns a valid JSON binary string.
+%% - decode/1 expects a valid JSON binary string, and returns erlang-formatted JSON.
+%%
+%% or a tuple {Mod, {encode, Opts}, {decode, Opts}} where Mod implements encode/2 and
```
seriyps: do you think it makes sense to add extra tags `encode` and `decode`? Won't just `{Mod, EncodeOpts, DecodeOpts}` be enough?

src/datatypes/epgsql_codec_json.erl
```diff
@@ -17,18 +17,47 @@
 
 -define(JSONB_VERSION_1, 1).
 
-%% TODO: JSON encode/decode `fun Mod:Name/1` / `{Mod, Name}` as option.
-%% Shall not pass `fun(_) -> .. end`, because of hot code upgrade problems.
-init(_, _) -> [].
+%% JsonMod shall either be a module with encode/1 and decode/1, where:
+%%
+%% - encode/1 expects erlang-formatted JSON, e.g. a map or a proplist,
+%%   and returns a valid JSON binary string.
```
seriyps: I don't think we should insist on requirement to return binary - iolist might be also fine (however, I don't know any erlang JSON library that returns iolist)

src/datatypes/epgsql_codec_json.erl
```diff
@@ -17,18 +17,47 @@
...
+%%      third-party JSON module in a new module that implements the required interface,
+%%      and pass in _that_ as JsonMod.
+init(JsonMod, _) ->
+    JsonMod.
 
 names() ->
     [json, jsonb].
 
+encode(ErlJson, json, JsonMod) when is_atom(JsonMod) ->
+    JsonMod:encode(ErlJson);
```
seriyps: I would expect complaints from elvis code linter about variable-module-calls from non-behaviour module. I think it makes sense to make `epgsql_codec_json` a behaviour module by defining 4 `-callback` attributes.

src/datatypes/epgsql_codec_json.erl
```diff
@@ -17,18 +17,47 @@
...
+encode(ErlJson, json, {JsonMod, {encode, Opts}, _}) when is_atom(JsonMod) ->
+    JsonMod:encode(ErlJson, Opts);
+encode(ErlJson, jsonb, JsonMod) when is_atom(JsonMod) ->
+    <<?JSONB_VERSION_1:8, (JsonMod:encode(ErlJson))/binary>>;
+encode(ErlJson, jsonb, {JsonMod, {encode, Opts}, _}) when is_atom(JsonMod) ->
+    <<?JSONB_VERSION_1:8, (JsonMod:encode(ErlJson, Opts))/binary>>;
```
seriyps: AFAIK, when you are doing `<<OneByte:8, HugeBinary/binary>>` whole `HugeBinary` is copied (I might be wrong). I would recommend using iolist here: `[?JSONB_VERSION_1 | JsonMod:encode(ErlJson, Opts)]` or `[<<?JSONB_VERSION_1:8>> | JsonMod:encode(ErlJson, Opts)]`

src/datatypes/epgsql_codec_json.erl
UmanShahzad: That'll definitely be better :+1:

src/datatypes/epgsql_codec_json.erl
UmanShahzad: Would that require the `JsonMod` implementing those callbacks to specify the `-behavior` directive too though? If so that'd effectively require every user of this feature to create their own mini wrapper module around their favorite JSON library to implement the callbacks.

src/datatypes/epgsql_codec_json.erl
UmanShahzad: I would be surprised but I also don't know. Likely there's just a new binary containing `OneByte` plus a reference to `HugeBinary`.

src/datatypes/epgsql_codec_json.erl
UmanShahzad: Sure!

src/datatypes/epgsql_codec_json.erl
UmanShahzad: Actually, reading around a bit, I think it might just create a completely new binary.

Let's use an iolist.

UmanShahzad:
@seriyps Thanks for the review; I'll address all of it soon. I'll add tests once we have an agreed upon implementation.

About the copying overhead - we can document it, but by default this won't do anything new, so users who're worried about performance can take this into their own hands and simply avoid the feature.

src/datatypes/epgsql_codec_json.erl
seriyps: no, it's NOT required for callback to actually have `-behavior` attribute

seriyps:
@UmanShahzad yep, this overhead is definitely not a blocker! Just good to be aware of it.

seriyps:
FYI, regarding appends of a big binaries. I made a small investigation and that's what I found:
TLDR: it does copy big binary.

Here are the minimal example:
```
$ cat bin_append.erl
-module(bin_append).

-export([main/1]).

main(HugeBin) ->
  <<7, HugeBin/binary>>.
```
And here it's erlang assembler and beam instructions:
```
$ cat bin_append.erl
-module(bin_append).

-export([main/1]).

main(HugeBin) ->
  <<7, HugeBin/binary>>.

$ erl
1> c(bin_append).
2> erts_debug:df(bin_append).

$ cat bin_append.dis
0007F398C4744D0: i_func_info_IaaI 0 bin_append main 1
00007F398C4744F8: move_x1_c 0
00007F398C474508: i_gc_bif1_jIsId j(0000000000000000) byte_size/1 x(0) 2 x(2)
00007F398C474538: bs_add_jssId j(0000000000000000) x(1) x(2) 1 x(1)
00007F398C474568: bs_add_jssId j(0000000000000000) x(1) 1 1 x(1)
00007F398C474598: i_bs_init_fail_xjId x(1) j(0000000000000000) 1 x(1)
00007F398C4745C0: bs_put_string_II 1 -1941485734
00007F398C4745D8: i_new_bs_put_binary_all_jsI j(0000000000000000) x(0) 8
00007F398C4745F8: move_return_x x(1)
<..>

$ erlc -S bin_append.erl
$ cat bin_append.S
{module, bin_append}.  %% version = 0

{exports, [{main,1},{module_info,0},{module_info,1}]}.

{attributes, []}.

{labels, 7}.

{function, main, 1, 2}.
  {label,1}.
    {line,[{location,"bin_append.erl",5}]}.
    {func_info,{atom,bin_append},{atom,main},1}.
  {label,2}.
    {line,[{location,"bin_append.erl",6}]}.
    {move,{integer,0},{x,1}}.
    {gc_bif,byte_size,{f,0},2,[{x,0}],{x,2}}.
    {bs_add,{f,0},[{x,1},{x,2},1],{x,1}}.
    {bs_add,{f,0},[{x,1},{integer,1},1],{x,1}}.
    {bs_init2,{f,0},{x,1},0,1,{field_flags,[]},{x,1}}.
    {bs_put_string,1,{string,[7]}}.
    {bs_put_binary,{f,0},{atom,all},8,{field_flags,[unsigned,big]},{x,0}}.
    {move,{x,1},{x,0}}.
    return.

<..>
```

so, we see following instructions sequence:

```
i_bs_init_fail_xjId
bs_put_string_II
i_new_bs_put_binary_all_jsI
```

1st one creates a new preallocated binary
2nd adds `7` to it
3rd adds big binary to it

Looking at how `i_new_bs_put_binary_all` is implemented:

https://github.com/erlang/otp/blob/113f6917a343a5db98e4771598f1a190779ca2e8/erts/emulator/beam/bs_instrs.tab#L302-L316

it calls `erts_new_bs_put_binary_all`

https://github.com/erlang/otp/blob/2a95d46dd626b0acc57e79b596d00dea01f8270a/erts/emulator/beam/erl_bits.c#L1009-L1040

which indeed copies the binary by calling `copy_binary_to_buffer`

seriyps:
Implementation looks good now! Thanks! Would be nice to add:

1. Some minimal test
2. Some notes in README, probably in the end of https://github.com/epgsql/epgsql#data-representation section
3. Might be nice to export some `-type` for `JsonMod` option, smth as simple as `{module, any(), any()}` will be ok. But it's minor.

UmanShahzad:
@seriyps Nice investigation! I'll need to look at that more closely soon :)

I've written a test; will add some more documentation/types.

UmanShahzad:
@seriyps This should be ready now.

UmanShahzad:
@davidw Hey David, can you check this out?

P.S. have been using it in my own code without any issues.

seriyps:
Merged. Thanks!

PR #200, opened 2019-10-01
Use ubuntu trusty for OTP 18 on Travis

larshesel:


seriyps:
Thanks!

larshesel:
I just noticed that this PR went to master instead of `devel` - sorry about that - the right target should be `devel`, right?

seriyps:
Ouch.. yep... would you mind creating one more pr to devel?

larshesel:
New one in #201.

PR #204, opened 2020-01-18
fix range codecs

MikhailKalashnikov:
Fix encode/decode for epgsql_codec_intrange and epgsql_codec_timerange.
Before this fix following selects end with a error:

```
epgsql:equery(Conn, "SELECT tstzrange(current_timestamp, null)").
epgsql:equery(Conn, "SELECT int4range(1, 1)").
```

### DIFF
--- a/src/datatypes/epgsql_codec_intrange.erl
+++ b/src/datatypes/epgsql_codec_intrange.erl
@@ -16,7 +16,7 @@
 
 -export_type([data/0]).
 
--type data() :: {left(), right()}.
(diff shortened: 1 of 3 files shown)


seriyps:
Thanks! Please, add tests. Failing build is not a problem (test dependency is not compatible with OTP18); can be fixed later.

MikhailKalashnikov:
@seriyps I added tests

seriyps:
Ok, merged! Thanks everyone!

PR #205, opened 2020-01-29
Add table metadata to #column{} record

seriyps:
PostgreSQL `RowDescription` message [contains](https://www.postgresql.org/docs/12/protocol-message-formats.html) information about result column's database table and column location.
This commit exposes this data as 2 new fields of `#column{}` record:
- table_oid
  If the field can be identified as a column of a specific table, the OID of the table; otherwise zero.
- table_attr_number
  If table_oid is not zero, the attribute number of the column; otherwise zero.

Thos fields are exposed by, eg, [Python's psycopg2](https://www.psycopg.org/docs/extensions.html#psycopg2.extensions.Column.table_oid).

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,21 +126,15 @@ Asynchronous connect example (applies to **epgsqli** too):
 ### Simple Query
 
 ```erlang
--type query() :: string() | iodata().
--type squery_row() :: {binary()}.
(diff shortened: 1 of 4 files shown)


seriyps:
Thanks!

PR #207, opened 2020-01-31
Make it possible to provide `#statement{}` to `epgsql:prepared_query/3`

seriyps:
So, there will be no need to call `describe` each time - less round-trips.

### DIFF
--- a/README.md
+++ b/README.md
@@ -283,25 +283,26 @@ squery including final `{C, Ref, done}`.
 ### Prepared Query
 
 ```erlang
-{ok, Columns, Rows}        = epgsql:prepared_query(C, StatementName, [Parameters]).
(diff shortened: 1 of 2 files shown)


davidw:
Any docs or tests needed?

seriyps:
README was updated to reflect this.
Tests - I'm not sure any new tests are needed, because of the way this change is implemented:

```
prepared_query(.., #statement{}, ..) ->
    epgsql_sock:call(...);
prepared_query(.., Name, ..) ->
   prepared_query(.., parse(Name), ..).
```

So, if one calls it "the old way", it will recurse through the "new" path. So, this new way of calling this function is already covered.

seriyps:
No problems! Thanks!

PR #208, opened 2020-01-31
Enable test-coverage

seriyps:
Fail the build if coverage falls below 55%

Would be nice to raise this bar in the future.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,8 @@ src/epgsql_errcodes.erl:
 	./generate_errcodes_src.sh > src/epgsql_errcodes.erl
 
 test: compile
-	@$(REBAR) do ct -v
+	@$(REBAR) do ct -v -c
+	@$(REBAR) cover -v -m 55
 
 dialyzer: compile
 	@$(REBAR) dialyzer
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -16,7 +16,8 @@ src/epgsql_errcodes.erl:
 	./generate_errcodes_src.sh > src/epgsql_errcodes.erl
 
 test: compile
-	@$(REBAR) do ct -v
+	@$(REBAR) do ct -v -c
+	@$(REBAR) cover -v -m 55
```
davidw: Could use a comment or maybe some notes somewhere explaining what this is: it's not obvious from a glance.

seriyps: Good idea! Will do.

seriyps: Made it a named constant. Also added eunit because we actualy have 2 unit-tests

seriyps: Or you mean to explain what `cover` is?

davidw: Something like

# Fail the build if coverage falls below 55%

seriyps:
I added the comment. Thanks for suggestion!

PR #209, opened 2020-01-31
Introduce epgsql{a,i}:execute_batch/3

seriyps:
It is useful when there is a need to execute the same query with a different sets of parameters.
The most obvious example is batch insert:

```erlang
RowsToInsert = [
  ["Joe", 35],
  ["Paul", 26],
  ["Mary", 24],
  ...
],
{_Cols, [{ok, [{1}]}, {ok, [{2}]} | _ ]} = 
  epgsql:execute_batch(
    C, "INSERT INTO account (name, age) VALUES ($1, $2) RETURNING id", 
    RowsToInsert)
```

The one controversal detail that I'm not sure about is when to return `[#columns{}]`? Because they are only needed when 2nd argument is sql query string. Otherwise they can be extracted from `Statement#statement.columns`.

There are options:

1. Don't return columns at all (in this case one who needs them will have to use `parse` + `execute_batch`)
2. Always return `{Cols, Results}` - as it is implemented in this PR
3. Return `{Cols, Results}` when 2nd argument is SQL string and return just `Results` if it's a `#statement{}`

### DIFF
--- a/README.md
+++ b/README.md
@@ -289,8 +289,8 @@ squery including final `{C, Ref, done}`.
 {error, Error}             = epgsql:prepared_query(C, "non_existent_query", [Parameters]).
 ```
 
-`Parameters` - optional list of values to be bound to `$1`, `$2`, `$3`, etc.
-`Statement` - name of query given with ```erlang epgsql:parse(C, StatementName, "select ...", []).```
+- `Parameters` - optional list of values to be bound to `$1`, `$2`, `$3`, etc.
+- `Statement` - name of query given with ```erlang epgsql:parse(C, StatementName, "select ...", []).```
                (can be empty string) or `#statement{}` record returned by `epgsql:parse`.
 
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -385,23 +385,54 @@ Batch execution is `bind` + `execute` for several prepared statements.
...
+```
+
+equivalent:
+
+```erlang
+epgsql:execute_batch(C, "INSERT INTO account (name, age) VALUES ($1, $2) RETURNING id",
+                     [ ["Joe", 35], ["Paul", 26], ["Mary", 24] ]).
+```
+
+In case one of the batch items will cause an error, result returned for this particular
```
davidw: "In case one of the batch items causes an error, the result returned ..."

seriyps: Thanks! Fixed.

seriyps:
Thank you!

PR #210, opened 2020-01-31
Re-generate epgsql_errcodes.erl (and update generator script)

seriyps:
Just synchronized error-codes file with the latest postgresql sources


seriyps:
Thanks!

Issue #203, opened 2019-11-06
Feature: Map null to undefined on decoding


Xelgun:
What do you think? Afaik the usage of null atom in Erlang code is much less common and also it's not Elixir-compatible.
We can't handle null to undefined conversion with pluggable datatype codecs because of this lines
https://github.com/epgsql/epgsql/blob/cd3144d88f476b0425ea42319a3f6a8a22c13edf/src/epgsql_wire.erl#L144-L149 (and few others) and also because NULL is not really a PostgreSQL data type.

In jiffy for example we have {null_term, Term} option that cause to return the specified Term instead of null on decoding.

seriyps:
Hi. Yep, I was thinking about this and also think it should be configurable for both encoding and decoding. I think some special option for `connect` would work the best. As alternative - it might be set via epgsql's application env.

Also, my idea was to allow multiple values for null encoding (eg, encode all of `undefined`, `null`, `nil` as `NULL`), but always return one of them when decoding. The simplest way this could be achieved is to provide accepted values as a list and use 1st element of a list when decoding. Eg:

```
epgsql:connect(#{..., null_terms => [undefined, null, nil]})
```

This connection will encode any of `undefined`, `null`, `nil` as NULL, but will decode all `NULL`s as `undefined` because it's 1st element of a list.

Unfortunately, it's not that straightforward to implement (would require separate field in `epgsql_sock`'s state, needs to be bypassed deep into decoder's code stack), so PR is welcome!

seriyps:
Ok, it looks like it might be simplified a bit if `null` handling will be moved from `epgsql_wire` to `epgsql_binary` and `null_terms` will be stored in `#codec.opts`. But still, it would require separate field in `epgsql_sock`s state (not a big problem)

seriyps:
With #212 it is configurable now.

PR #212, opened 2020-02-04
Make it possible to configure NULL representation.

### LINKED ISSUE #193: null to text conversion in array
Hello, suppose I have an array of string where some elements are nulls. The result of select of the array is:
```
epgsql:equery(C, "SELECT $1", [[<<"s1">>, null]]).        
{error,{error,error,<<"42P18">>,indeterminate_datatype,
              <<"could not determine data type of parameter $1">>,
              [{file,<<"postgres.c">>},
               {line,<<"1336">>},
               {routine,<<"exec_parse_message">>},
               {severity,<<"ERROR">>}]}}
```
cast fixes above error
```
epgsql:equery(C, "SELECT $1::TEXT[]", [[<<"s1">>, null]]).
{ok,[{column,<<"text">>,{array,text},1009,-1,-1,1}],
    [{[<<"s1">>,<<"null">>]}]}

```
but in the result the null value is also converted to string. How to avoid above error  and preserve type of null values?

seriyps:
New `nulls` `connect/1` option added that can be used to configure the set of terms to be used to represent NULL in input (parameters) and output (data rows).

Also, NULLs inside arrays are fixed #193

My original intention was to fix #193, but I found that it would be easier to add configurable nulls #203 at the same time.

It might affect performance slightly (because we are using `lists:member/2` instead of function clause match in a quite hot path, but the difference should be proportional to the size of `nulls` lists and it's up to user to decide on trade-offs. If all the elements of `nulls` are atoms, `lists:member` should be quite quick.

Implementation is not super clean, because there are many places where we do NULL encoding/decoding: `record` type, `array` type and generic case. Probably can be refactored later. But all the cases are covered by tests.

Some usage examples are in tests: `epgsql_SUITE:custom_null`


src/epgsql_binary.erl
```diff
@@ -213,44 +245,73 @@ decode_record1(<<Oid:?int32, Len:?int32, ValueBin:Len/binary, Rest/binary>>, Siz
...
-    {Data, {NDims, Lengths}} = encode_array(Array, 0, [], ValueEncoder),
+encode_array(Array, Oid, ArrayEncoder) ->
+    {Data, {NDims, Lengths, HasNull}} = encode_array_dims(Array, ArrayEncoder),
     Lens = [<<N:?int32, 1:?int32>> || N <- lists:reverse(Lengths)],
-    Hdr  = <<NDims:?int32, 0:?int32, Oid:?int32>>,
+    HasNullInt = case HasNull of
+                     true -> 1;
+                     false -> 0
+                 end,
+    Hdr  = <<NDims:?int32, HasNullInt:?int32, Oid:?int32>>,
```
seriyps: before this change we [always](https://github.com/epgsql/epgsql/commit/688a55dc0c1faf6eb93f252e81478269f150b7b8) had this `HasNullInt` flag set to `0` which violates PostgreSQL server expectations, but seems server still was able to handle that.

https://github.com/postgres/postgres/blob/177be9edf4bb966400db7769d61e479aa0fe0201/src/backend/utils/adt/arrayfuncs.c#L1434

src/epgsql_wire.erl
```diff
@@ -246,16 +246,18 @@ encode_parameters([P | T], Count, Formats, Values, Codec) ->
...
-encode_parameter({T, undefined}, Codec) ->
-    encode_parameter({T, null}, Codec);
-encode_parameter({_, null}, _Codec) ->
-    {1, <<-1:?int32>>};
-encode_parameter({{unknown_oid, _Oid}, Value}, _Codec) ->
-    {0, encode_text(Value)};
 encode_parameter({Type, Value}, Codec) ->
-    {1, epgsql_binary:encode(Type, Value, Codec)};
-encode_parameter(Value, _Codec) ->
-    {0, encode_text(Value)}.
```
seriyps: this clause was unreachable because `Value` is always 2-tuple

src/epgsql_binary.erl
seriyps: Seems this flag is validated but then ignored by server: https://github.com/postgres/postgres/blob/177be9edf4bb966400db7769d61e479aa0fe0201/src/backend/utils/adt/arrayfuncs.c#L1304-L1308

davidw:
You might add a brief comment in the README like

"As documented below, if this parameter is not passed in, the default is `null`"

seriyps:
Done. Thanks!

PR #216, opened 2020-03-04
response type in epgsql_cmd_(e)query is not only {error, #error{}} but can also be {error, sock_closed}

onno-vos-dev:
Dialyzer on one of my projects is unhappy when I handle this particular kind of error

### DIFF
--- a/src/commands/epgsql_cmd_equery.erl
+++ b/src/commands/epgsql_cmd_equery.erl
@@ -24,7 +24,7 @@
 -type response() :: {ok, Count :: non_neg_integer(), Cols :: [epgsql:column()], Rows :: [tuple()]}
                   | {ok, Count :: non_neg_integer()}
                   | {ok, Cols :: [epgsql:column()], Rows :: [tuple()]}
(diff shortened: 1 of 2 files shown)


seriyps:
Hi. Given that `sock_closed` is not specific to any particular command (any command can fail with this reason because it's generated not by command itself, but by `epgsql_sock`), I would suggest to update `epgsql:equery` / `epgsql:squery` typespecs to be smth like

```erlang
-spec equery(connection(), sql_query(), [bind_param()]) ->
                    epgsql_cmd_equery:response() | {error, sock_closed}.
```

Or even create a separate type `epgsql_sock:error()` (it can include `sync_required`, `closed`, `sock_closed`, probably smth else - grep for `flush_queue`) and do

```
erlang
-spec equery(connection(), sql_query(), [bind_param()]) ->
                    epgsql_cmd_equery:response() | epgsql_sock:error().
```

onno-vos-dev:
Closing in favor of https://github.com/epgsql/epgsql/pull/221 due to the branch name on this PR no longer describing the change correctly.

PR #223, opened 2020-03-05
Do not re-use the same connection timeout during different steps of establishing the connection

enidgjoleka:
Hi,

I was looking through the `epgsql:connect/x` and stumbled upon the connection_timeout. 
`epgsql:connect/x` accepts "timeout" as a property under the `Opts`. 
However, this property is not respected while trying to establish a connection under `epgsql_cmd_connect:execute/2` since this function calls `gen_tcp:connect` and `ssl:connect` with the same Timeout. In the worst case scenario we might end up with almost 2*Timeout as our connection timeout.

This PR tries to address this issue.
Please let me know if I missed something.

### DIFF
--- a/src/commands/epgsql_cmd_connect.erl
+++ b/src/commands/epgsql_cmd_connect.erl
@@ -124,9 +124,8 @@ maybe_ssl(S, false, _, PgSock, _Deadline) ->
 maybe_ssl(S, Flag, Opts, PgSock, Deadline) ->
     ok = gen_tcp:send(S, <<8:?int32, 80877103:?int32>>),
     Timeout0 = timeout(Deadline),
-    {ok, <<Code>>} = gen_tcp:recv(S, 1, Timeout0),
-    case Code of
-        $S  ->
+    case gen_tcp:recv(S, 1, Timeout0) of
+        {ok, <<$S>>}  ->
             SslOpts = maps:get(ssl_opts, Opts, []),
             Timeout = timeout(Deadline),
             case ssl:connect(S, SslOpts, Timeout) of
@@ -136,13 +135,15 @@ maybe_ssl(S, Flag, Opts, PgSock, Deadline) ->
                     Err = {ssl_negotiation_failed, Reason},
                     {error, Err}
             end;
-        $N ->
+        {ok, <<$N>>} ->
             case Flag of
                 true ->
                     epgsql_sock:set_net_socket(gen_tcp, S, PgSock);
                 required ->
                     {error, ssl_not_available}
-            end
+            end;
+        {error, Reason} ->
+            {error, Reason}
     end.
 
 %% Auth sub-protocol
@@ -280,12 +281,7 @@ hex(Bin) ->
     <<<<(HChar(H)), (HChar(L))>> || <<H:4, L:4>> <= Bin>>.
 
 deadline(Timeout) ->
-    erlang:monotonic_time(millisecond) + Timeout.
+    erlang:monotonic_time(milli_seconds) + Timeout.
 
 timeout(Deadline) ->
-    case (Deadline - erlang:monotonic_time(millisecond)) of
-        NewDeadline when NewDeadline < 0 ->
-            0;
-        Val ->
-            Val
-    end.
+    erlang:max(0, Deadline - erlang:monotonic_time(milli_seconds)).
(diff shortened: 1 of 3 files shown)


enidgjoleka:
Actually I got the inspiration for using `inet:*_timer` from `gen_tcp:connect`.
If you are concerned about using `inet` undocumented functions I can as well use the respective: `erlang:start_timer/x, erlang:read_timer/x; erlang:cancel_timer`.

seriyps:
But I don't think `erlang:start_timer` will help if you are doing blocking calls? Say

```
erlang:start_timer(.., 1000),
gen_tcp:connect(..., 10000)
```

it will block for 10s, not 1s… Or you mean something else?

enidgjoleka:
My intention is like following:
1. Timer = erlang:start_timer(.., Timeout),
2. gen_tcp:connect(..., Timeout),
3. Timeout2 = erlang:read_timer(Timer),
4. ssl:connect(..., Timeout2)

The important points in here are:
* We should start the timer with the same timeout like the first network call.
* If the network call takes longer than the timeout it is its responsibility meaning `gen_tcp:connect` to time out properly.
* If gen_tcp does not time out then we read the new remaining timeout and pass it to the last network call.
* It is again the responsibility of `ssl:connect` to timeout given the new timeout.

seriyps:
Ah, I see, so, more or less the same as https://github.com/epgsql/epgsql/pull/223#pullrequestreview-370271822 but using `erlang:*_timer`. Should be fine as well (but you should take care of timer's `{timeout, ...}` message I guess?). But please try to craft some test for it and handle situation when `erlang:read_timer(T)` returns `false`.

enidgjoleka:
@seriyps I went with your proposal instead of using erlang:*_timer since it looked more neat.

src/commands/epgsql_cmd_connect.erl
```diff
@@ -119,15 +119,16 @@ hide_password(PasswordFun) when is_function(PasswordFun, 0) ->
...
-maybe_ssl(S, false, _, PgSock) ->
+maybe_ssl(S, false, _, PgSock, _Deadline) ->
     epgsql_sock:set_net_socket(gen_tcp, S, PgSock);
-maybe_ssl(S, Flag, Opts, PgSock) ->
+maybe_ssl(S, Flag, Opts, PgSock, Deadline) ->
     ok = gen_tcp:send(S, <<8:?int32, 80877103:?int32>>),
-    Timeout = maps:get(timeout, Opts, 5000),
-    {ok, <<Code>>} = gen_tcp:recv(S, 1, Timeout),
+    Timeout0 = timeout(Deadline),
+    {ok, <<Code>>} = gen_tcp:recv(S, 1, Timeout0),
```
seriyps: would be cool to move this inside a case and handle error in case like
```erlang
case gen_tcp:recv(...) of
  {ok, $S} ->
    …;
  {ok, $N} ->
    …;
  {error, …} ->
    …
end
```

I know `recv` errors were not handled before, but now the probability of getting timeout from it becomes higher.

src/commands/epgsql_cmd_connect.erl
```diff
@@ -277,3 +278,14 @@ hex(Bin) ->
                (N) when N < 16 -> $W + N
             end,
     <<<<(HChar(H)), (HChar(L))>> || <<H:4, L:4>> <= Bin>>.
+
+deadline(Timeout) ->
+    erlang:monotonic_time(millisecond) + Timeout.
+
+timeout(Deadline) ->
+    case (Deadline - erlang:monotonic_time(millisecond)) of
```
seriyps:
```suggestion
        erlang:max(0, Deadline - erlang:monotonic_time(millisecond))
```

seriyps: Just noticed CI failure on OTP-18. Seems we'll need to use deprecated `milli_seconds` time unit.

src/commands/epgsql_cmd_connect.erl
enidgjoleka: Will fix this!

src/commands/epgsql_cmd_connect.erl
enidgjoleka: Will fix this!

seriyps:
Thanks @enidgjoleka and @davidw !

PR #226, opened 2020-03-16
Use application name when establishing a connection

enidgjoleka:
Hi,

This PR tries to introduce `application_name` as an optional parameter that can be used while establishing the connection. The name will be displayed in the pg_stat_activity view and included in CSV log entries.
You can read here for more info: https://www.postgresql.org/docs/9.6/libpq-connect.html#LIBPQ-CONNSTRING


seriyps:
Should we consider usage of some generic parameter with a list of K/V pairs to send with `StartupMessage`? I'm not really sure how this mechanism works.

enidgjoleka:
I was looking for such list and I barely managed to find the application_name in the link that I posted. My inspiration for the application_name came from cockroachdb and as of now I can not think of another parameter that is missing.

seriyps:
I found this 2:

https://github.com/postgres/postgres/blob/113758155c11cf993ca0ecee8856e300a2525a30/src/interfaces/libpq/fe-protocol3.c#L2175-L2202

https://github.com/postgres/postgres/blob/113758155c11cf993ca0ecee8856e300a2525a30/src/interfaces/libpq/fe-connect.c#L362-L378

So, seems the list of options is limited and not that big. I think it's ok to go with specific (non-generic) parameters.

BTW, could you consider generalizing the handling of parameters (all those "user", "database",  "replication", "application_name")?

Smth like

```erlang
Params = [
  {database, "database"},
  {application_name, "application_name"},
  ...
],
lists:foldl(
  fun({Key, Param}, ConnStr) ->
    case maps:find(Key, Opts) of
         error -> ConnStr;
         {ok, Vak}  -> ["database", 0, Database, 0 | ConnStr]
    end
  end, ["user", 0, Username, 0], Params)
```

But in general I'm ok to merge as is. Will see if @davidw have any objections.

seriyps:
Than you!

PR #227, opened 2020-03-20
Handle tls connection during cancellation

enidgjoleka:
Hi,

Looking at the `epgsql:cancel/x` it appears like `epgsql` opens a new temporary TCP connection and sends the cancellation request. This is fine for applications which uses normal non-TLS connections. 

However, if we are dealing with an application which uses TLS connections for all its queries except for the cancellation request then we might open ourselves to the man-in-the-middle-attack.
The attacker can in theory listen to our cancellation request and detect the previously generated secret key during the connection establishment of the original connection.
Once the attacker gets hold of that key while the original connection can still accept queries even after its last query was cancelled then the attacker can in theory cancel all the queries fired from the original connection.

This PR tries to address this issue.
I decided to rewrite part of `epgsql_cmd_connect` to better decouple opening a socket with the rest of the logic. This newly introduced function, `open_socket/x`, was then called from `epgsql_sock.erl` during the cancellation. 

Please let me know if I missed something!

### DIFF
--- a/src/commands/epgsql_cmd_connect.erl
+++ b/src/commands/epgsql_cmd_connect.erl
@@ -6,7 +6,7 @@
 %%%
 -module(epgsql_cmd_connect).
 -behaviour(epgsql_command).
--export([hide_password/1, opts_hide_password/1]).
+-export([hide_password/1, opts_hide_password/1, open_socket/2]).
 -export([init/1, execute/2, handle_message/4]).
 -export_type([response/0, connect_error/0]).
 
@@ -47,14 +47,37 @@
 init(#{host := _, username := _} = Opts) ->
     #connect{opts = Opts}.
 
-execute(PgSock, #connect{opts = #{host := Host} = Opts, stage = connect} = State) ->
-    Timeout = maps:get(timeout, Opts, 5000),
-    Deadline = deadline(Timeout),
-    Port = maps:get(port, Opts, 5432),
+execute(PgSock, #connect{opts = #{username := Username} = Opts, stage = connect} = State) ->
     SockOpts = [{active, false}, {packet, raw}, binary, {nodelay, true}, {keepalive, true}],
-    case gen_tcp:connect(Host, Port, SockOpts, Timeout) of
-        {ok, Sock} ->
-            client_handshake(Sock, PgSock, State, Deadline);
+    epgsql_sock:set_attr(connect_opts, Opts, PgSock),
+    case open_socket(SockOpts, Opts) of
+        {ok, Mode, Sock} ->
+            PgSock1 = epgsql_sock:set_net_socket(Mode, Sock, PgSock),
+            Opts2 = ["user", 0, Username, 0],
+            Opts3 = case maps:find(database, Opts) of
+                        error -> Opts2;
+                        {ok, Database}  -> [Opts2 | ["database", 0, Database, 0]]
+                    end,
+           {Opts4, PgSock2} =
+               case Opts of
+                   #{replication := Replication}  ->
+                       {[Opts3 | ["replication", 0, Replication, 0]],
+                        epgsql_sock:init_replication_state(PgSock1)};
+                   _ -> {Opts3, PgSock1}
+               end,
+            Opts5 = case Opts of
+                        #{application_name := ApplicationName}  ->
+                            [Opts4 | ["application_name", 0, ApplicationName, 0]];
+                        _ ->
+                            Opts4
+                    end,
+           ok = epgsql_sock:send(PgSock2, [<<196608:?int32>>, Opts5, 0]),
+           PgSock3 = case Opts of
+                         #{async := Async} ->
+                             epgsql_sock:set_attr(async, Async, PgSock2);
+                         _ -> PgSock2
+                     end,
+           {ok, PgSock3, State#connect{stage = maybe_auth}};
         {error, Reason} = Error ->
             {stop, Reason, Error, PgSock}
     end;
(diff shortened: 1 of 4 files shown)


src/commands/epgsql_cmd_connect.erl
```diff
@@ -47,22 +47,52 @@
...
-execute(PgSock, #connect{opts = #{host := Host} = Opts, stage = connect} = State) ->
-    Timeout = maps:get(timeout, Opts, 5000),
-    Deadline = deadline(Timeout),
-    Port = maps:get(port, Opts, 5432),
+execute(PgSock, #connect{opts = #{username := Username} = Opts, stage = connect} = State) ->
     SockOpts = [{active, false}, {packet, raw}, binary, {nodelay, true}, {keepalive, true}],
-    case gen_tcp:connect(Host, Port, SockOpts, Timeout) of
-        {ok, Sock} ->
-            client_handshake(Sock, PgSock, State, Deadline);
+    PgSock1 = epgsql_sock:set_attr(connect_opts, Opts, PgSock),
```
seriyps: Opts includes `password`, which is a `fun()`. If you will store it in a permanent state it will crash during code upgrade. I think we should filter and store only necessary options there.

src/commands/epgsql_cmd_connect.erl
```diff
@@ -47,22 +47,59 @@
...
+           {Opts4, PgSock3} =
+               case Opts of
+                   #{replication := Replication}  ->
+                       {[Opts3 | ["replication", 0, Replication, 0]],
+                        epgsql_sock:init_replication_state(PgSock2)};
+                   _ -> {Opts3, PgSock2}
+               end,
+            Opts5 = case Opts of
+                        #{application_name := ApplicationName}  ->
+                            [Opts4 | ["application_name", 0, ApplicationName, 0]];
```
enidgjoleka: On my previous PR, https://github.com/epgsql/epgsql/pull/226, I introduced a bug in here by returning `Opts3` rather than `Opts4`. I fixed it in here.

test/epgsql_SUITE.erl
```diff
@@ -284,6 +296,67 @@ connect_with_ssl(Config) ->
...
+              , {timeout, 1000} ]
+            ],
+    {ok, C} = Module:connect(Host, "epgsql_test", Args2),
+    ?assertMatch({ok, _Cols, [{true}]},
+                Module:equery(C, "select ssl_is_used()")),
+    process_flag(trap_exit, true),
+    Self = self(),
+    spawn_link(fun() ->
+                   ?assertMatch(?QUERY_CANCELED, Module:equery(C, "SELECT pg_sleep(5)")),
+                   Self ! done
```
seriyps: another option is to send the result to `Self` and assert in receiver:

```erlang
Self ! {done, Module:equery(C, "..")}
end,
<..>
receive {done, Result} ->
    ?assertMatch(?QUERY_CANCELLED, Result)
after ....
```

test/epgsql_SUITE.erl
```diff
@@ -284,6 +296,67 @@ connect_with_ssl(Config) ->
...
+    {Host, Port} = epgsql_ct:connection_data(Config),
+    Module = ?config(module, Config),
+    Args2 = [ {port, Port}, {database, "epgsql_test_db1"}
+            | [ {ssl, true}
+              , {timeout, 1000} ]
+            ],
+    {ok, C} = Module:connect(Host, "epgsql_test", Args2),
+    ?assertMatch({ok, _Cols, [{true}]},
+                Module:equery(C, "select ssl_is_used()")),
+    process_flag(trap_exit, true),
```
seriyps: do we really need this `trap_exit`, given that we don't try to receive `{'EXIT'` messages? I think CT case will fail if it's killed by linked process

src/epgsql_sock.erl
```diff
@@ -225,17 +228,20 @@ handle_cast(stop, State) ->
...
     Msg = <<16:?int32, 80877102:?int32, Pid:?int32, Key:?int32>>,
-    ok = gen_tcp:send(Sock, Msg),
-    gen_tcp:close(Sock),
+    case epgsql_cmd_connect:open_socket(SockOpts, ConnectOpts) of
+      {ok, Mode, Sock} when Mode == gen_tcp ->
+          ok = gen_tcp:send(Sock, Msg),
+          gen_tcp:close(Sock);
+      {ok, Mode, Sock} when Mode == ssl ->
+          ok = ssl:send(Sock, Msg),
+          ssl:close(Sock);
```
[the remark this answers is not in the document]
enidgjoleka: It makes perfect sense. Will merge the two {ok...} clauses into one.

test/epgsql_SUITE.erl
enidgjoleka: I will remove it.

src/commands/epgsql_cmd_connect.erl
```diff
@@ -47,22 +47,59 @@
...
+                        _ ->
+                            Opts4
+                    end,
+           ok = epgsql_sock:send(PgSock2, [<<196608:?int32>>, Opts5, 0]),
```
seriyps: I think you should use `PgSock3` here

src/epgsql_sock.erl
```diff
@@ -225,17 +228,17 @@ handle_cast(stop, State) ->
...
     SockOpts = [{active, false}, {packet, raw}, binary],
-    %% TODO timeout
-    {ok, Sock} = gen_tcp:connect(Addr, Port, SockOpts),
     Msg = <<16:?int32, 80877102:?int32, Pid:?int32, Key:?int32>>,
-    ok = gen_tcp:send(Sock, Msg),
-    gen_tcp:close(Sock),
+    case epgsql_cmd_connect:open_socket(SockOpts, ConnectOpts) of
+      {ok, Mode, Sock} ->
+          ok = apply(Mode, send, [Sock, Msg]),
+          apply(Mode, close, [Sock]);
```
seriyps: I don't know if current erlang compiler can optimize this, but in general `Mode:close(Sock)` performs better than `apply(Mode, close, [Sock])` because arguments are transferred via registers and not as list. But it's minor, difference is negligible.

enidgjoleka: Originally I went for `Mode:close` but Elvis had opinions on it, so I reverted back to `apply`.

enidgjoleka:
It would be nice if I could get some traction on this PR.

src/commands/epgsql_cmd_connect.erl
seriyps: so, how about this one?

seriyps:
Merged! Sorry for the slow response.

PR #228, opened 2020-04-01
Handle skipped commands in execute_batch

enidgjoleka:
This PR tries to address the problem raised by #186 .
Whenever `epgsql/i/a:execute_batch` is being called with several statements/sql_queries and PG returns an error for one statement then epgsql would return an error on unexpected message 90 ("Z").
The only exception to this rule is if you call `epgsql/i/a:execute_batch` with a list of three statements where the first one is successful, the second crashes and the third will be skipped. But had it been a fourth statement it would have crashed.

With the changes introduced in this PR epgsql won't return an error unexpected_message but rather a list of type [ {ok, result()}, {error, [ {error, #error{}}, {error, skipped} ] } ].
There will be as many {error, skipped} as many statements were not executed from PG.

### DIFF
--- a/src/commands/epgsql_cmd_batch.erl
+++ b/src/commands/epgsql_cmd_batch.erl
@@ -36,7 +36,8 @@
 -type response() :: [{ok, Count :: non_neg_integer(), Rows :: [tuple()]}
                      | {ok, Count :: non_neg_integer()}
                      | {ok, Rows :: [tuple()]}
-                     | {error, epgsql:query_error()}].
+                     | {error, [{error, epgsql:query_error()} | {error, skipped}]}
+                     ].
 -type state() :: #batch{}.
 
 -spec init(arguments()) -> state().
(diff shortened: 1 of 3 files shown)


src/commands/epgsql_cmd_batch.erl
```diff
@@ -110,13 +111,14 @@ handle_message(?COMMAND_COMPLETE, Bin, Sock,
...
     {add_result, Result, {complete, Complete}, Sock, State#batch{batch = Batch}};
-handle_message(?READY_FOR_QUERY, _Status, Sock, #batch{batch = B} = _State) when
-      length(B) =< 1 ->
+handle_message(?READY_FOR_QUERY, _Status, Sock, _State) ->
     Results = epgsql_sock:get_results(Sock),
     {finish, Results, done, Sock};
 handle_message(?ERROR, Error, Sock, #batch{batch = [_ | Batch]} = State) ->
     Result = {error, Error},
-    {add_result, Result, Result, Sock, State#batch{batch = Batch}};
+    Skipped = lists:duplicate(length(Batch), {error, skipped}),
```
seriyps, line 119: Well, I don't really like the proposed result format (that we are generating some fake error structures).
I think what we should do is to just explicitly document that "in case of error in one of the items of the batch, all the remaining queries of that batch will be ignored. So, last element of result list will be `{error, #error{}}` and the length of the result list might be shorter that the length of the batch." + some small example.

I think the simplest way to achieve that is to just to remove the `when length(B) =< 1` guard and that's it.
Can also add some consistency check like:

```erlang
handle_message(?READY_FOR_QUERY, _Status, Sock, #batch{batch = B}) ->
  Results = epgsql_sock:get_results(Sock),
  case B of
    [] -> ok;
   _ ->
    % make sure the last result is `{error, #error{}}`, because it should be the only case when
    % we receive ?READY_FOR_QUERY before all batch items are processed
    {error, _} = lists:last(Results)
 end,
 {finish, ...}.
```
but it's very much optional. It should not be reachable unless there is some bug in the server

enidgjoleka: Looking at this with fresh eyes I agree with you.
Will simply remove the guard and fix the docs.

seriyps:
Merged! Thanks @enidgjoleka @davidw !

PR #229, opened 2020-04-26
Add sasl_prep implementation for validating passwords according to sa…

enidgjoleka:
…sl specification

### DIFF
--- a/src/epgsql_sasl_prep_profile.erl
+++ b/src/epgsql_sasl_prep_profile.erl
@@ -1,0 +1,145 @@
+%%% coding: utf-8
+%%% @doc
+%%% sal_prep profile helper module that will validate a utf-8
+%%% string based on sasl_prep profile as defined in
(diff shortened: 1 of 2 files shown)


src/epgsql_sasl_prep_profile.erl
```diff
@@ -0,0 +1,159 @@
...
+
+-module(epgsql_sasl_prep_profile).
+
+-export([ validate/1
+        ]).
+
+-spec validate(iolist()) -> iolist().
+validate(Str) ->
+    lists:foreach(fun(F) ->
+                      lists:any(F, unicode:characters_to_list(Str, utf8))
```
seriyps, line 16: Please move `unicode:characters_to_list(Str, utf8)` out of inner loop.

seriyps:
Ok, merging now. Thanks @enidgjoleka

Issue #225, opened 2020-03-10
Rework the travis pipeline


onno-vos-dev:
I'd like to take a stab at re-working the travis pipeline and do the following work: 

- Use docker for Postgres 
  - Use different versions of Postgres (which will increase how long the pipeline takes)
  - Using Docker will also allow contributors to run tests locally

- Even though .travis.yml specifies `trusty` (Ubuntu 14.04) it seems that Travis decided that Xenial is a better distro (magic or am I missing something?) I'd like to bump it to `Bionic Beaver` to ensure that the LTS isn't dropped next year. 

- As long as it won't explode the diff outside of .travis.yml, I'd like to try and include [CockroachDB](http://cockroachlabs.com/) in the pipeline considering that any Erlang project that intends to use or try CockroachDB will end up relying on epgsql.

If you think it's a good idea and if you have any other suggestions, please let me know :+1:

seriyps:
* I'm ok with docker if you really want to invest your time in it. There is already an issue for that #127 
  * It would be interesting to test on different versions, I'm agree. I don't think it will increase time if builds are run in parallel
  *  I personaly don't see any problems installing postgres on my computer, but I see there is a demand. Maybe depends on OS?
* Sure. Just be aware that not all OTP versions were available on all Ubuntu versions. See #201 
  * Yep, the reason was that OTP18 was not available on any other Ubuntu on Travis. That affected quite many erlang projects: https://github.com/for-GET/jesse/pull/83 , https://github.com/gen-smtp/gen_smtp/pull/179
* I saw a note on cockroach in my email, but now I see it was removed. But I think it's a good idea. But we will definitely won't be able to run all the same testcases on Cockroach since there are many features that are different (eg, smaler choice of datatypes, no replication protocol support on Cockroach). I guess it might be achieved with some smart common-test suite grouping technique?

onno-vos-dev:
There are a few reasons why I'd like to invest the time in using docker:
- Local testing is only possible by installing Postgres locally. Something that yes, is not a big deal but also should not be necessary for testing purposes. 
- It falls under the cockroachdb part since that could easily be tested in Docker version "you can install cockroach locally". Most developers will be interested in one or the other, the ability of spinning up a docker to test either or, or both, is a big gain. I indeed removed it and intended to open another issue for it but figured I'd start with this task. It's something that I have in the back of my mind and can link to this issue (and #127) as a dependency. 

[...]

PR #237, opened 2020-07-18
test with otp 23 in travis

galdor:
Since OTP 23 is the last stable Erlang version, it makes sense to test it.

I'm more concerned about the fact that CI only uses PostgreSQL 10, while both 11 and 12 are available and used everywhere. I'm not sure on how to test multiple Pg versions in Travis though.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ install: "true"
 language: erlang
 matrix:
   include:
+    - otp_release: 23.0
     - otp_release: 22.2
     - otp_release: 21.3
     - otp_release: 20.3


seriyps:
Cool, thanks!

Regarding Postgresql: it should be possible to test with different erlang and PG versions using Travis "build matrix". But last time I checked Travis was really slow in providing support for a newer versions of PostgreSQL. But if you have any insights on how to do this it would be cool!

seriyps:
Also, there exists this ticket from colleague of mine: https://github.com/epgsql/epgsql/issues/225
I have to check if he's going to work on (he's on vacation now).

PR #239, opened 2020-07-18
switch to crypto:mac/4 since crypto:hmac/3 is deprecated

### LINKED ISSUE #235: Use of deprecated function (crypto:hmac/3)
When building epgsql on OTP 23, I get:

```_build/default/lib/epgsql/src/epgsql_scram.erl:129: Warning: crypto:hmac/3 is deprecated and will be removed in OTP 24; use use crypto:mac/4 instead```

I'd like to submit a patch to switch to `crypto:/mac/4`, but this function has only been introduced in OTP 22.1, and I'm not sure which versions epgsql officially supports, and I can't seem to find how to precisely check for the OTP version with the preprocessor.

Would it be ok to drop <22.1 ? If not, do you have a suggestion on how to detect the version at compile time ?

galdor:
close #235

It could be made cleaner and simpler by dropping support for OTP<21 (which introduced the `OTP_RELEASE` macro if I'm not mistaken). I fail to see the point in supporting releases that old.

### DIFF
--- a/src/epgsql_scram.erl
+++ b/src/epgsql_scram.erl
@@ -116,8 +116,18 @@ hi1(Str, U, Hi, I) ->
     Hi1 = bin_xor(Hi, U2),
     hi1(Str, U2, Hi1, I - 1).
 
+-ifdef(OTP_RELEASE).
+-if(OTP_RELEASE >= 23).
+hmac(Key, Str) ->
+    crypto:mac(hmac, sha256, Key, Str).
+-else.
 hmac(Key, Str) ->
     crypto:hmac(sha256, Key, Str).
+-endif.
+-else.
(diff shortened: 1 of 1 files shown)


seriyps:
Thanks.
Yep, ufortunately it's not that clean and beautiful solution (I can't think of a better one though), but function is small enough so the amount of duplicaion is minimal.

We prefer to not drop old OTP releases unless it requires significant effort to support (eg, OTP < 17 had no maps support and maps in OTP < 18 were much slower so we dropped those releases quite early). I know there exists a plenty of big systems that depend on epgsql which are still supported, but running on an old OTP versions because of too much effort to migrate. They can use older epgsql for sure, but I think it's better to "keep the doors open" as long as effort is low.

Issue #137, opened 2017-08-30
Add support for COPY


seriyps:
Protocol description
https://www.postgresql.org/docs/current/static/protocol-flow.html#PROTOCOL-COPY
Protocol packet description
https://www.postgresql.org/docs/current/static/protocol-message-formats.html

We should also think on API:

Should we accept `file:io_device()`? Just raw binary() / iolist() data? Filenames `file:name_all()`.

davidw:
Is there a use case driving this, or is it just a general "we should do this"?

seriyps:
I personally don't have use cases for it, but I just noticed that most of the PG drivers supports this. Eg,
elixir: https://hexdocs.pm/postgrex/Postgrex.html#stream/4
Python: http://initd.org/psycopg/docs/usage.html#using-copy-to-and-copy-from
Rust: https://docs.rs/postgres/0.15.1/postgres/stmt/struct.Statement.html#method.copy_in
NodeJS: https://github.com/brianc/node-pg-copy-streams
Go (partialy): https://github.com/lib/pq#future--things-you-can-help-with

So yep, it's more close to "why not?"

GuillaumeGandon:
> Is there a use case driving this, or is it just a general "we should do this"?

From [PostgreSQL documentation](https://www.postgresql.org/docs/current/populate.html) it is the fastest method to populate a DB:

1. > Note that loading a large number of rows using COPY is **almost always faster** than using INSERT, even if PREPARE is used and multiple insertions are batched into a single transaction.

1. > If you are loading a freshly created table, the **fastest method** is to create the table, bulk load the table's data using COPY, then create any indexes needed for the table.

dszoboszlay:
> Is there a use case driving this, or is it just a general "we should do this"?

I have a very specific use case for this feature, as I'd like to convert a lot of data from a different DB to Postgres.

> We should also think on API:

The DB conversion goes like: read a chunk of data from one DB, maybe make some schema adjustments in memory and then dump to Postgres. This would need some streaming API from `epgsql`, because it's not feasible to hold the entire DB to be converted in memory. Dumping it to disk and reading back is a waste of time too.

It would be also useful if `epgsql` could do the conversion between Erlang types and PG types as usual.

seriyps:
I was thinking about implementing it as an Erlang's [io protocol](https://erlang.org/doc/apps/stdlib/io_protocol.html).

So, smth like
```erlang
[...]

PR #248, opened 2021-01-07
COPY .. FROM STDIN

seriyps:
All 3 formats are supported: text, CSV, binary. Partially solves #137.

For text-based formats the way you send COPY payload is to epgsql is by using Erlang IO-protocol (so, standard library functions like `io:put_chars`, `io:format/3`, `file:write` can be used).

For binary format a special function was introduced: `epgsql:copy_send_rows/3`. It takes care of encoding Erlang values to postgresql binary format. It should be possible to use IO-protocol to transfer binary format as well (so, encoding is done bu the user), but I haven't provided any helpers or tests for that.

A lot of examples you can find in the tests, new epgsql_copy_SUITE test suite was introduced.

I'm still not sure if the `epgsql:copy_from_stdin` function should take raw SQL query OR take some map with parameters and construct SQL inside. Chosen raw SQL for now because it's easier and more flexible, but it's also a bit more easy to break when, eg, number of columns or format do not match expectations - I added a lots of guards to catch such situations.
Anyway, still open for discussions on API, and that's the reason why I haven't added those new functions to README yet - let's consider them undocumented and experimental for now. I'll add more documentation later in a separate PR.

In https://github.com/epgsql/epgsql/issues/137#issuecomment-744498403 I said that it might be a problem that we use `inet` driver directly instead of `gen_tcp:send`, but I later realized that it's not that a big problem, because `erlang:port_command` is actually blocking when socket buffers are full.

I would like to add `COPY .. TO STDOUT` some time later, but don't have any exact plans yet. We might need to implement some TCP flow control before, because, with combination of `{active, true}` and large COPY result set we can easily end up with OOM.

### DIFF
--- a/src/commands/epgsql_cmd_copy_done.erl
+++ b/src/commands/epgsql_cmd_copy_done.erl
@@ -15,11 +15,13 @@
 
 %% -include("epgsql.hrl").
 -include("protocol.hrl").
+-include("../epgsql_copy.hrl").
 
 init(_) ->
     [].
 
 execute(Sock0, St) ->
+    #copy{} = epgsql_sock:get_subproto_state(Sock0), % assert we are in copy-mode
     {PktType, PktData} = epgsql_wire:encode_copy_done(),
     Sock1 = epgsql_sock:set_packet_handler(on_message, Sock0),
     Sock = epgsql_sock:set_attr(subproto_state, undefined, Sock1),
--- a/src/commands/epgsql_cmd_copy_done.erl
+++ b/src/commands/epgsql_cmd_copy_done.erl
@@ -1,0 +1,42 @@
+%%% @doc Tells server that the transfer of COPY data is done.
+%%%
+%%% It makes server to "commit" the data to the table and switch to the normal command-processing
+%%% mode.
+%%%
+%%% @see epgsql_cmd_copy_from_stdin
+
+-module(epgsql_cmd_copy_done).
+-behaviour(epgsql_command).
+-export([init/1, execute/2, handle_message/4]).
+-export_type([response/0]).
+
+-type response() :: {ok, Count :: non_neg_integer()}
+                  | {error, epgsql:query_error()}.
+
+%% -include("epgsql.hrl").
+-include("protocol.hrl").
+
+init(_) ->
+    [].
+
+execute(Sock0, St) ->
+    {PktType, PktData} = epgsql_wire:encode_copy_done(),
+    Sock1 = epgsql_sock:set_packet_handler(on_message, Sock0),
+    Sock = epgsql_sock:set_attr(subproto_state, undefined, Sock1),
+    {send, PktType, PktData, Sock, St}.
+
+handle_message(?COMMAND_COMPLETE, Bin, Sock, St) ->
+    Complete = epgsql_wire:decode_complete(Bin),
+    Res = case Complete of
+        {copy, Count} -> {ok, Count};
+        copy -> ok
+    end,
+    {add_result, Res, {complete, Complete}, Sock, St};
+handle_message(?ERROR, Error, Sock, St) ->
+    Result = {error, Error},
+    {add_result, Result, Result, Sock, St};
+handle_message(?READY_FOR_QUERY, _Status, Sock, _State) ->
+    [Result] = epgsql_sock:get_results(Sock),
+    {finish, Result, done, Sock};
+handle_message(_, _, _, _) ->
+    unknown.
--- a/src/commands/epgsql_cmd_copy_done.erl
+++ b/src/commands/epgsql_cmd_copy_done.erl
@@ -35,12 +35,8 @@ execute(Sock0, St) ->
     end.
 
 handle_message(?COMMAND_COMPLETE, Bin, Sock, St) ->
-    Complete = epgsql_wire:decode_complete(Bin),
-    Res = case Complete of
-        {copy, Count} -> {ok, Count};
-        copy -> ok
-    end,
-    {add_result, Res, {complete, Complete}, Sock, St};
+    Complete = {copy, Count} = epgsql_wire:decode_complete(Bin),
+    {add_result, {ok, Count}, {complete, Complete}, Sock, St};
 handle_message(?ERROR, Error, Sock, St) ->
     Result = {error, Error},
     {add_result, Result, Result, Sock, St};
--- a/src/commands/epgsql_cmd_copy_done.erl
+++ b/src/commands/epgsql_cmd_copy_done.erl
@@ -21,11 +21,18 @@ init(_) ->
     [].
 
 execute(Sock0, St) ->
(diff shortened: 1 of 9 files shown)


davidw (approved):

seriyps:
I rebased over #248 to fix the build and also fixed pre-OTP-21 compatibility `catch T:R:S` syntax

src/commands/epgsql_cmd_copy_done.erl
```diff
@@ -0,0 +1,51 @@
...
+            Pkts = [{?COPY_DATA, epgsql_wire:encode_copy_trailer()},
+                    {PktType, PktData}],
+            {send_multi, Pkts, Sock, St}
+    end.
+
+handle_message(?COMMAND_COMPLETE, Bin, Sock, St) ->
+    Complete = epgsql_wire:decode_complete(Bin),
+    Res = case Complete of
+        {copy, Count} -> {ok, Count};
+        copy -> ok
```
fcristovao, line 41: This seems like it's never a possibility from `epgsql_wire:decode_complete`, nor it would fulfill the `response()` type above.

src/commands/epgsql_cmd_copy_from_stdin.erl
```diff
@@ -0,0 +1,103 @@
...
+%%% > CopyData*            -- implemented in io protocol, not here
+%%% > CopyDone | CopyFail  -- implemented in epgsql_cmd_copy_done
+%%% < CommandComplete      -- implemented in epgsql_cmd_copy_done
+%%% '''
+-module(epgsql_cmd_copy_from_stdin).
+-behaviour(epgsql_command).
+-export([init/1, execute/2, handle_message/4]).
+-export_type([response/0]).
+
+-type response() :: ok | {error, epgsql:query_error()}.
```
fcristovao, line 35: This type doesn't match the response in `handle_message(?COPY_IN_RESPONSE, ... `.
It should be `{ok, list(text | binary)}` or maybe something even stricter.

As I was looking through the tests though, I also wondered, what is the point of returning a list of `text | binary`? It's confusing from a API's standpoint.
Since the values will always have to be the same, I think it would be best to return just `text | binary`, instead of a list of them.
Additionally, I think it would also be good to return a `csv` case, as they are distinct formats.

src/commands/epgsql_cmd_copy_from_stdin.erl
```diff
@@ -0,0 +1,103 @@
...
+                          {?COPY_DATA, Header}], Sock, St}
+    end.
+
+%% CopyBothResponses
+handle_message(?COPY_IN_RESPONSE, <<BinOrText, NumColumns:?int16, Formats/binary>>, Sock,
+               #copy_stdin{initiator = Initiator, format = RequestedFormat}) ->
+    ColumnFormats =
+        [case Format of
+             0 -> text;
+             1 -> binary
```
fcristovao, line 67: nit: consider turning this into a simple "translator" function.
Then you can also use it for `init_copy_state(someFun(BinOrText), ...`.
Notice the last function clause on `init_copy_state/4` will actually error out with `{format_mismatch, 0, binary}` instead of `{format_mismatch, text, binary}`.

src/epgsql_wire.erl
```diff
@@ -310,6 +316,39 @@ encode_standby_status_update(ReceivedLSN, FlushedLSN, AppliedLSN) ->
...
+encode_copy_row(Values, Types, Codec) ->
+    NumCols = length(Types),
+    [<<NumCols:?int16>>
+    | [
+       case epgsql_binary:is_null(Value, Codec) of
+           true ->
+               <<-1:?int32>>;
+           false ->
+               epgsql_binary:encode(Type, Value, Codec)
+       end || {Type, Value} <- lists:zip(Types, Values) % TODO: parallel iteration ninstead
```
fcristovao, line 343: nit: `ninstead` -> `instead`

src/epgsql_wire.erl
```diff
@@ -310,6 +316,39 @@ encode_standby_status_update(ReceivedLSN, FlushedLSN, AppliedLSN) ->
...
+      ]].
```
fcristovao, line 344: nittest of nits: consider optimizing the `lists:zip` away by recursively calling a function and matching on the heads of Values and Types to produce these.

src/commands/epgsql_cmd_copy_from_stdin.erl
seriyps: > It should be {ok, list(text | binary)} or maybe something even stricter.

Good catch

> what is the point of returning a list of text | binary?

Yeah, one reason might be because it says how many columns are expected in the input data. Another reason is because it's possible in the future that they will be different (arguable). And finally, it's because I added all the asserts some time after I was done with definition of return values :smiley:

> good to return a csv case

I'd say, epgsql doesn't see any difference between text and csv (it only appears in SQL query and epgsql normally doesn't look inside SQL queries). But we can, eg, make 3rd argument of `copy_from_stdin` accept `text | csv | {binary, ...}` and use it instead. Not sure if it worth this complication.

src/epgsql_wire.erl
seriyps: Yeah.. Also, even better optimization could be achieved by implementation of a function similar to `epgsql_wire:build_decoder`. Because right now we do a lots of internal "codec" table lookups on each data "cell" and it can be cached, because each data row has the same structure. This optimization could be reused by `epgsql:execute_batch` as well.
I'm thinking about making a separate PR with all those optimizations.

src/commands/epgsql_cmd_copy_done.erl
seriyps: Yeah, it is an artifact from me changing line 220 in epgsql_wire much later after I added this snippet. Seems it should be relatively safe to assume it always returns `COPY 1234` and never just `COPY`:

> CommandComplete (B)
> <..>
> For a COPY command, the tag is COPY rows where rows is the number of rows copied. (Note: the row count appears only in PostgreSQL 8.2 and later.)

https://www.postgresql.org/docs/current/protocol-message-formats.html

> On successful completion, a COPY command returns a command tag of the form
> 
> COPY count
> The count is the number of rows copied.

https://www.postgresql.org/docs/current/sql-copy.html#id-1.9.3.55.7

So I think it's safe as long as we don't support PostgreSQL older than 8.2 (postgresql.org says the oldest supported postgres server version is 9.5; but I think I heard that AWS Redshift uses older version?)

src/epgsql_wire.erl
seriyps: Just accidentally discovered there exists `lists:zipwith/3`. So, Can be rewritten as

```suggestion
| lists:zipwith(
    fun(Type, Value) ->
        case epgsql_binary:is_null(Value, Codec) of
           true ->
               <<-1:?int32>>;
           false ->
               epgsql_binary:encode(Type, Value, Codec)
       end
    end, Types, Values)].
```

fcristovao (approved):

seriyps:
Thanks for thew reviews!

PR #249, opened 2021-01-07
Stick to rebar3_lint 0.2.0

seriyps:
Version 0.3.0 exists, but it has some troubles working from non-default profile


seriyps:
I think I'll just merge it. It's quite minor and it will unblock the build for #248

PR #250, opened 2021-01-15
Optimize batch encoder

seriyps:
Before the only way to encode Erlang term to PostgreSQL binary format was to use `epgsql_wire:encode_parameters` that calls `epgsql_binary:encode/3` in a loop. The problem is that `epgsql_binary:encode/3` does quite a lot of `codec()` database lookups and structure transformations and this process repeats for each parameter.
It worked quite ok for normal `equery`, because you usually don't send too many parameters to it. But now with introduction of `COPY .. FROM STDIN WITH (FORMAT binary)` we might need to encode a lot of rows of the same structure. So it makes sense to do all the OID table lookups in advance.
The same optimization was applied to `execute_batch/3`, because we use the same statement for multiple rows.

1st commit optimizes COPY, 2nd optimizes `execute_batch`

### DIFF
--- a/src/commands/epgsql_cmd_batch.erl
+++ b/src/commands/epgsql_cmd_batch.erl
@@ -48,6 +48,7 @@ init(Batch) when is_list(Batch) ->
     #batch{batch = Batch}.
 
 execute(Sock, #batch{batch = Batch, statement = undefined} = State) ->
+    %% Each query has it's own statement
     Codec = epgsql_sock:get_codec(Sock),
     Commands =
         lists:foldr(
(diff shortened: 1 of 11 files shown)


src/commands/epgsql_cmd_batch.erl
```diff
@@ -48,6 +48,7 @@ init(Batch) when is_list(Batch) ->
     #batch{batch = Batch}.
 
 execute(Sock, #batch{batch = Batch, statement = undefined} = State) ->
+    %% Each query has it's own statement
```
davidw, line 51: Linguistic nitpick: "its", not "it's". "It's" == "it is"

davidw (approved):

seriyps:
I'm not merging yet, because I plan to wait for my colleagues to do some benchmarks before.

seriyps:
Ok, we tried to import some multi-terabyte dump using both "current" and "optimized" version, but haven't noticed any difference, primarily because the import happened to be disk-bound :smile: . At least, it haven't made anything worse.
I'll either try to add some microbenchmarks or merge it as is bit later.

daironm-hillrom:
Is this ready to merge?

PR #252, opened 2021-01-27
Add support for timeout query

enidgjoleka:
This PR tries to achieve the same outcome as https://github.com/epgsql/epgsql/pull/129

However, it takes a different approach to the whole client side timeout implementation:
* It does not rely on gen_server's timeout mechanism nor does it use internal timers in `epgsql_sock` due to the potential race conditions that may occour. @seriyps has described some of those race conditions on the original PR.
* Achieves hard client side timeouts by leveraging `epsqla` API functions for delivering the full results combined with waiting for the results via `receive.... after Timeout end`.  If the timeout is reached the connection process will be stopped and `{error, timeout}` will be returned to the caller.

Hard timeouts could have been implemented via spawning a new process that handles the interaction with epgsql_sock gen_server and kill both of them in the event of a timeout. However, this approach would have meant more data to be copied over between processes which may become an issue for queries returning large volumes of data.
Hence, the  idea of @seriyps  to use `epqslqa` combined with `receive...after Timeout end`


enidgjoleka:
It would be great to get some traction here.

src/epgsql.erl
```diff
@@ -546,3 +596,19 @@ to_map(Map) when is_map(Map) ->
...
+
+-spec wait_for_result(connection(), reference(), timeout()) ->
+    epgsql_cmd_squery:response() | epgsql_sock:error() | {error, timeout}.
+wait_for_result(C, Ref, Timeout) ->
+    receive
+        {C, Ref, Result} -> Result
+    after Timeout ->
+        epgsql_sock:kill(C),
+        {error, timeout}
+    end.
```
seriyps, line 608: I think there is a potential problem / race conditions with this implementation of the timeout (which can be avoided by using query cancellation with the price of being non-hard timeout):

After we call `epgsql_sock:kill`, there is a chance that the query that "timed out" would still complete successfully

https://swimlanes.io/d/-O2ywUSdJ

Say, you call `epgsql:squery(C, "COMMIT", 5000)`. epgsql_sock sends "COMMIT" packet to the server, server executes COMMIT for 5 seconds and sends "COMAND COMPLETE" to epgsql, epgsql maybe receives it and starts parsing. Exactly at the same time `after` timer of `receive` is triggered, so `wait_for_result` casts `kill` and immediately returns `{error, timeout}`.

a) `epgsql_sock` may receive `COMMAND COMPLETE` from TCP just before it receives `kill`. Then it would send the (unexpected) `{epgsql, ref(), ok}` response back to the caller and also transaction would be successfully committed
b) `epgsql_sock` may receive `kill` while `COMMAND COMPLETE` is on the way from postgresql to `epgsql_sock`. Then there will be no stale `{epgsql, ref(), ok}` message, but there is a chance that transaction will be committed even though we returned `{error, timeout}`.

There might be any other command in place of `COMMIT`. Say, someone might execute `INSERT` or `UPDATE` or some DDL command outside of transaction... It's less of a problem for `SELECT` of course.

By closing the connection "abruptly", we don't have any synchronization point, so, we can't say if the request have completed or have been rolled-back.

This problem would be even more severe for `execute_batch`, because it would have multiple independent commands in a pipeline, some of which may complete and some not before the timeout.

This race-condition should be less problematic when every modifying operation is executed inside of transaction (so, only `COMMIT` might be affected by race-condition). But not everyone are so strict about transactions, so, it might be risky to introduce the generic API with potential race-condition.

Another thing to note is that this whole timeout API was built entirely on top of existing epgsql APIs (well, except newly-added `kill` command). So, maybe it's better to have those commands be implemented outside of `epgsql` module (say, add `epgsqlt` - note `t` in the end for "timeout", module, or have it as a separate app) and only have core functionality in the `epgsql` module?

Also, I think this potential race-condition should be mentioned in the README. If we think race-condition is not a problem, then some explanation why.

seriyps: I think I might need to try to create some script/tool that tries to simulate this race condition. Say, have a table with single counter field and hammer it with smth like

```
epgsql:squery(C, "SELECT pg_sleep(0.001 * X); UPDATE my_tab SET count = count+1;", X)
```

then count the number of `ok` and `{error, timeout}` responses and compare it with `count` value in the table.

test/epgsql_SUITE.erl
```diff
@@ -459,6 +463,12 @@ select(Config) ->
         [{<<"1">>, <<"one">>}, {<<"2">>, <<"two">>}] = Rows
       end).
 
+squery_with_timeout(Config) ->
+    F1 = fun(C) -> ?assertMatch({error, timeout}, epgsql:squery(C, "SELECT pg_sleep(2)", 2000)) end,
```
seriyps, line 467: isn't it potentially flaky? Maybe use at least `1500` as a last argument?

README.md
```diff
@@ -288,7 +288,7 @@ end.
 ```
 
 - `Statement` - parsed statement (see parse below)
-- `Res` has same format as return value of `epgsql:equery/3`.
+- `Res` has same the format as the return value of `epgsql:equery/3`.
```
seriyps, line 291: nit:

```diff
- has same the format
+ has the same format
```

PR #255, opened 2021-02-26
Fix `-if(?OTP_RELEASE)` (was missing question mark)

seriyps:
Found out that `-if` works not as `-if(MACRO_NAME..)` but as `-if(?MACRO_NAME ...)`. So, epgsql crashes on OTP-24 release candidate


davidw (approved):

seriyps:
thanks!

PR #257, opened 2021-04-07
Do not format certain attributes of State on abnormal epgsql sock termination

enidgjoleka:
Imagine the following scenario:
We have fired an `epgsql:equery/squery` call and while getting the results back from PG server we are accumulating rows in the `epgsql_sock` state. While this is happening `epgsql_sock` process terminates abnormally. 
At this moment gen_server will try to print/format the State.
Given that we might have accumulated several hundred thousands or even million of rows in the state this will cause a memory spike and in the worst case scenario an OOM.

This solution tries to solve that problem by introducing another gen_server optional callback, `format_status`, where we replace the accumulated list in the rows with an atom stating `information_redacted`.

While looking at the code I realized that `format_status` could be used to remove the sensitive info  from `connect_opts` that we are currently hiding via lambdas.

### DIFF
--- a/src/epgsql_sock.erl
+++ b/src/epgsql_sock.erl
@@ -54,7 +54,7 @@
          copy_send_rows/3,
          standby_status_update/3]).
 
--export([handle_call/3, handle_cast/2, handle_info/2]).
+-export([handle_call/3, handle_cast/2, handle_info/2, format_status/2]).
 -export([init/1, code_change/3, terminate/2]).
 
 %% loop callback
@@ -95,7 +95,7 @@
                 current_cmd_transport :: transport() | undefined,
                 async :: undefined | atom() | pid(),
                 parameters = [] :: [{Key :: binary(), Value :: binary()}],
-                rows = [] :: [tuple()],
+                rows = [] :: [tuple()] | information_redacted,
                 results = [],
                 sync_required :: boolean() | undefined,
                 txstatus :: byte() | undefined,  % $I | $T | $E,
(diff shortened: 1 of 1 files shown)


seriyps:
Thanks!
Regarding password masking, I'm afraid it still has a chance to show up in crash logs as "last message in was ...", so, I'd prefer to leave it as is =)
Otherwise, looks good and makes a lot of sense

enidgjoleka:
You are right, password shows up as last message. 
I dropped the second commit for now :+1:

seriyps (approved):
Let's see if anyone else have smth to say, otherwise will merge in a couple of days.

src/epgsql_sock.erl
```diff
@@ -298,6 +298,13 @@ terminate(_Reason, #state{mod = ssl, sock = Sock}) -> ssl:close(Sock).
 code_change(_OldVsn, State, _Extra) ->
     {ok, State}.
 
+format_status(normal, [_PDict, State=#state{}]) ->
+  [{data, [{"State", State}]}];
+format_status(terminate, [_PDict, State]) ->
+  %% Do not format the rows attribute when process terminates abnormally
+  %% but allow it when is a sys:get_status/1.2
+  State#state{rows = information_redacted}.
```
davidw, line 306: In the shell, there's some kind of thing to print a term... but only up to a certain size. I forget what it is, but maybe something like that could be used to print a bit, but truncate it?

enidgjoleka: We introduced a similar tuning in our system that limits the size up to 50-100 but we did not have this in place for many many years until the last incident.
Hence, let's try to be defensive and not assume all the systems that use `epgsql` have this kind of tuning in place.

anha0825 (approved):

seriyps:
Sorry, forgot about it. Merged, thanks!

PR #258, opened 2021-05-17
Add OTP-24 to CI

seriyps:

PR #259, opened 2021-05-24
Make dialyzer warn about use of unknown types; fix some occurences of such

seriyps:
Also, updated the travis config a bit, since #258 is going to be more difficult than originally expected.


seriyps:
I think I'll merge it since it doesn't change any behavior.

PR #262, opened 2021-08-20
Migrate from travis to github actions, add support for OTP-24

seriyps:
It installs postgresql via `apt` because our common test helpers starts it's own postgresql server, so, it needs access to postgres binaries, not just already launched server.

It might be interesting to try to test on more PG versions, we could use ubuntu-20.04 for that because it supports multiple newer PG versions, but https://github.com/erlef/setup-beam action only supports OTP 20+ on this OS.
Another option is to add official PostgreSQL APT repository https://www.postgresql.org/download/linux/ubuntu/ and install postgresql server from there.
But that's for the next step. Now priority is to have at least some working CI


seriyps:
Ah, and the proof that it works:
https://github.com/seriyps/epgsql/pull/3
https://github.com/seriyps/epgsql/actions/runs/1152285963

anha0825 (approved):
Are you removing the travis file in a later PR?

seriyps:
@anha0825 Looks like I'm removing it in this PR already: https://github.com/epgsql/epgsql/pull/262/files#diff-6ac3f79fc25d95cd1e3d51da53a4b21b939437392578a35ae8cd6d5366ca5485

anha0825:
> @anha0825 Looks like I'm removing it in this PR already: https://github.com/epgsql/epgsql/pull/262/files#diff-6ac3f79fc25d95cd1e3d51da53a4b21b939437392578a35ae8cd6d5366ca5485

Ah, I missed that :)

seriyps:
Ok, I think I'm going to merge it. It seems to work fine from my fork..

PR #268, opened 2021-12-27
feat: support prepared query without pre-parsed

tigercl:


seriyps:
Hi. I'll have another look next week (on vacation right now), but first quick feedback:

1. I think better name would be smth like "cached_equery"
2. Right now the caching strategy is quite non-flexible. Might it be better to store the statements in some "external" cache (provided as callback/module) instead of map in the connection state? So users of the API could implement LRU / TTL etc.
3. Please add some description to the README
4. Please add tests
5. One of the tests should try to run two different sql queries with the same name, so it is clear what would be the error in that case. Maybe it's better to use SQL query itself (or it's hash) as cache key, so such type of error is not possible?

src/commands/epgsql_cmd_prepared_query2.erl
```diff
@@ -0,0 +1,144 @@
...
+%% So, statement can be reused multiple times.
+%% ```
+%% > Bind
+%% < BindComplete
+%% > Execute
+%% < DataRow*
+%% < CommandComplete
+%% > Sync
+%% < ReadyForQuery
+%% '''
```
seriyps, line 12: This doc loocs like plain copy-paste from `prepared_query`. please, adjust it with what the code actually does

tigercl:
Hi. Regarding the second point, I am not sure whether the same statement can be used by multiple connections at the same time? Can it really be like this? I think the statement is related to the connection, so we need to re-parse every time we reconnect to get the statement.

Regarding the fifth point, I haven't got a deep understanding of the current test cases, but I do plan to use SQL itself as the Name.

I will adjust the other points as soon as possible.

PR #270, opened 2022-03-03
Add `get_os_pid/1` function

seriyps:
It's similar to `SELECT pg_backend_pid()`, but does not need network roundtrip and can be called even when thefre is an asynchronous query is still running.

### DIFF
--- a/src/epgsql.erl
+++ b/src/epgsql.erl
@@ -12,7 +12,7 @@
          get_parameter/2,
          set_notice_receiver/2,
          get_cmd_status/1,
-         get_os_pid/1,
+         get_backend_pid/1,
          squery/2,
          equery/2, equery/3, equery/4,
          prepared_query/3,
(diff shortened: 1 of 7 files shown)


jesperes (approved):

src/epgsql.erl
```diff
@@ -260,6 +261,13 @@ set_notice_receiver(C, PidOrName) ->
 get_cmd_status(C) ->
     epgsql_sock:get_cmd_status(C).
 
+%% @doc Returns the OS pid of PostgreSQL backend OS process that serves this connection.
+%%
+%% Similar to `SELECT pg_get_pid()', but does not need network roundtrips.
+-spec get_os_pid(connection()) -> integer().
+get_os_pid(C) ->
```
enidgjoleka, line 268: I find it confusing the terminology around `os` since this call will be executed from a process within the BEAM that is potentially living in another machine compared to the PG server.
How about renaming it to `pg_backend_pid`?

seriyps: Good idea. I'd prefer `get_backend_pid` though, because `epgsql` is already about postgres

enidgjoleka (approved):

PR #275, opened 2022-12-21
A flow control in the replication streaming mode

featalion:
This implements support of `{active, N}` where `N ∈ [-32768, 32767]` on TCP and SSL (since OTP 21.3) sockets.
It makes possible to limit incoming data flow and activate it again on-demand.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,11 +108,9 @@ Only `host` and `username` are mandatory, but most likely you would need `databa
 - `ssl` if set to `true`, perform an attempt to connect in ssl mode, but continue unencrypted
   if encryption isn't supported by server. if set to `required` connection will fail if encryption
   is not available.
-- `ssl_opts` will be passed as is to `ssl:connect/3`. The `active` option is only available in
-  the `replication` mode and OTP >= 21.3.
+- `ssl_opts` will be passed as is to `ssl:connect/3`.
 - `tcp_opts` will be passed as is to `gen_tcp:connect/3`. Some options are forbidden, such as
-  `mode`, `packet`, `header`. The `active` option is only available in the `replication` mode.
-  When `tcp_opts` is not provided, epgsql does some tuning
+  `mode`, `packet`, `header`, `active`. When `tcp_opts` is not provided, epgsql does some tuning
   (eg, sets TCP `keepalive` and auto-tunes `buffer`), but when `tcp_opts` is provided, no
   additional tweaks are added by epgsql itself, other than necessary ones (`active`, `packet` and `mode`).
 - `async` see [Server notifications](#server-notifications)
@@ -127,6 +125,10 @@ Only `host` and `username` are mandatory, but most likely you would need `databa
 - `application_name` is an optional string parameter. It is usually set by an application upon
    connection to the server. The name will be displayed in the `pg_stat_activity`
    view and included in CSV log entries.
+- `socket_active` is an optional parameter, which can be true or integer in the range -32768
+  to 32767 (inclusive). This option only works in the replication mode and is used to control
+  the flow of incoming messages. See [Streaming replication protocol](#streaming-replication-protocol)
+  for more details.
 
 Options may be passed as proplist or as map with the same key names.
 
--- a/README.md
+++ b/README.md
@@ -108,9 +108,11 @@ Only `host` and `username` are mandatory, but most likely you would need `databa
 - `ssl` if set to `true`, perform an attempt to connect in ssl mode, but continue unencrypted
   if encryption isn't supported by server. if set to `required` connection will fail if encryption
   is not available.
-- `ssl_opts` will be passed as is to `ssl:connect/3`
+- `ssl_opts` will be passed as is to `ssl:connect/3`. The `active` option is only available in
+  the `replication` mode and OTP >= 21.3.
 - `tcp_opts` will be passed as is to `gen_tcp:connect/3`. Some options are forbidden, such as
-  `mode`, `packet`, `header`, `active`. When `tcp_opts` is not provided, epgsql does some tuning
+  `mode`, `packet`, `header`. The `active` option is only available in the `replication` mode.
+  When `tcp_opts` is not provided, epgsql does some tuning
   (eg, sets TCP `keepalive` and auto-tunes `buffer`), but when `tcp_opts` is provided, no
   additional tweaks are added by epgsql itself, other than necessary ones (`active`, `packet` and `mode`).
 - `async` see [Server notifications](#server-notifications)
--- a/doc/streaming.md
+++ b/doc/streaming.md
@@ -173,7 +173,5 @@ epgsql:activate(Connection).
 The `active` parameter of the socket will be set to the same value as it was configured in
 the connection's options.
 
-When the connection is in the synchronous mode, a provided callback module must implement
-`handle_socket_passive/1` function, which receives a current callback state and should
-return `{ok, NewCallbackState}`. The callback should not call `epgsql:activate/1` directly
-because it results in a deadlock.
+In the case of synchronous handler for replication messages `epgsql` will handle `socket_passive`
+messages internally.
--- a/doc/streaming.md
+++ b/doc/streaming.md
@@ -139,3 +139,38 @@ handle_x_log_data(StartLSN, EndLSN, Data, CbState) ->
     
     Otherwise (if you do not load all data from tables during erlang app startup) 
     it is not recommended to set align_lsn to true. In this case to stop PG server stop epgsql replication first.
+    
+## Flow control
+
+It is possible to set `{active, N}` on a TCP or SSL (since OTP 21.3) socket. E.g. for SSL:
+```erlang
+Opts = #{ host => "localhost"
+        , username => "me"
+        , password => "pwd"
+        , database => "test"
+        , ssl => true
+        , ssl_opts => [{active, 10}]
+        },
+{ok, Conn} = epgsql:connect(Opts).
+```
+
+It is currently allowed only in the replication mode. Its main purpose is to control the flow of
+replication messages from Postgresql database. If a database is under a high load and a process, which
+handles the message stream, cannot keep up with it then setting this option gives the handling process
+ability to get messages on-demand.
+
+When the connection is in the asynchronous mode, a process which owns a connection will receive
+```erlang
+{epgsql, Connection, socket_passive}
+```
+The process decides when to activate connection's socket again. To do that it should call:
+```erlang
+epgsql:activate(Connection).
+```
+The `active` parameter of the socket will be set to the same value as it was configured in
+the connection's options.
+
+When the connection is in the synchronous mode, a provided callback module must implement
+`socket_passive/1` function, which receives a current callback state and should
+return `{ok, NewCallbackState}`. The callback should not call `epgsql:activate/1` directly
+because it results in a deadlock.
--- a/doc/streaming.md
+++ b/doc/streaming.md
@@ -171,6 +171,6 @@ The `active` parameter of the socket will be set to the same value as it was con
 the connection's options.
 
 When the connection is in the synchronous mode, a provided callback module must implement
-`socket_passive/1` function, which receives a current callback state and should
+`handle_socket_passive/1` function, which receives a current callback state and should
 return `{ok, NewCallbackState}`. The callback should not call `epgsql:activate/1` directly
 because it results in a deadlock.
--- a/doc/streaming.md
+++ b/doc/streaming.md
@@ -142,14 +142,15 @@ handle_x_log_data(StartLSN, EndLSN, Data, CbState) ->
     
 ## Flow control
 
-It is possible to set `{active, N}` on a TCP or SSL (since OTP 21.3) socket. E.g. for SSL:
+It is possible to set `{socket_active, N}` on a [TCP](https://www.erlang.org/doc/man/inet.html#setopts-2)
+or [SSL](https://www.erlang.org/doc/man/ssl.html#setopts-2) (since OTP 21.3) socket. E.g. for SSL:
 ```erlang
-Opts = #{ host => "localhost"
-        , username => "me"
-        , password => "pwd"
-        , database => "test"
-        , ssl => true
-        , ssl_opts => [{active, 10}]
+Opts = #{host => "localhost",
+         username => "me",
+         password => "pwd",
+         database => "test",
+         ssl => require,
+         socket_active => 10
         },
 {ok, Conn} = epgsql:connect(Opts).
 ```
@@ -163,6 +164,8 @@ When the connection is in the asynchronous mode, a process which owns a connecti
 ```erlang
 {epgsql, Connection, socket_passive}
 ```
+as soon as underlying socket received N messages from network.
+
 The process decides when to activate connection's socket again. To do that it should call:
 ```erlang
 epgsql:activate(Connection).
--- a/src/epgsql.erl
+++ b/src/epgsql.erl
@@ -47,7 +47,7 @@
 -export_type([connection/0, connect_option/0, connect_opts/0,
               connect_error/0, query_error/0, sql_query/0, column/0,
               type_name/0, epgsql_type/0, statement/0,
-              transaction_option/0, transaction_opts/0]).
+              transaction_option/0, transaction_opts/0, socket_active/0]).
 
 %% Deprecated types
 -export_type([bind_param/0, typed_param/0,
@@ -66,6 +66,7 @@
 -type host() :: inet:ip_address() | inet:hostname().
 -type password() :: string() | iodata() | fun( () -> iodata() ).
 -type connection() :: pid().
+-type socket_active() :: true | -32768..32767.
 -type connect_option() ::
     {host, host()}                                 |
     {username, string()}                           |
@@ -80,7 +81,8 @@
     {codecs,   Codecs     :: [{epgsql_codec:codec_mod(), any()}]} |
     {nulls,    Nulls      :: [any(), ...]} |    % terms to be used as NULL
     {replication, Replication :: string()} | % Pass "database" to connect in replication mode
-    {application_name, ApplicationName :: string()}.
+    {application_name, ApplicationName :: string()} |
+    {socket_active, Active :: socket_active()}.
 
 -type connect_opts() ::
         [connect_option()]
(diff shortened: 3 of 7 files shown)


README.md
```diff
@@ -108,9 +108,11 @@ Only `host` and `username` are mandatory, but most likely you would need `databa
 - `ssl` if set to `true`, perform an attempt to connect in ssl mode, but continue unencrypted
   if encryption isn't supported by server. if set to `required` connection will fail if encryption
   is not available.
-- `ssl_opts` will be passed as is to `ssl:connect/3`
+- `ssl_opts` will be passed as is to `ssl:connect/3`. The `active` option is only available in
+  the `replication` mode and OTP >= 21.3.
 - `tcp_opts` will be passed as is to `gen_tcp:connect/3`. Some options are forbidden, such as
-  `mode`, `packet`, `header`, `active`. When `tcp_opts` is not provided, epgsql does some tuning
+  `mode`, `packet`, `header`. The `active` option is only available in the `replication` mode.
```
seriyps, line 114: I was thinking that `active` should be passed not directly to `ssl_opts`/`tcp_opts` but rather on the top-level of `epgsql:connect_opts()`, like `epgsql:connect(#{host => "..", username => "..", password => "..", socket_active => 64 | true})`. Just because all the other socket options are passed as-is to the underlying socket, but `active` is interpreted by `epgsql`. So later `epgsql_cmd_connect` and `epgsql_sock` would take the top-level `socket_active` config to set the `ssl`/`gen_tcp` `active` option based on this top-level config value.

It would also make options lookup faster, because it is just `maps:get(socket_active, ConnectOpts, true)` instead of proplist lookup in `update_active_opt` (it is important because it might be a quite hot code path when N is small)

doc/streaming.md
```diff
@@ -138,4 +138,39 @@ handle_x_log_data(StartLSN, EndLSN, Data, CbState) ->
     then you do not have any risk to lose data. 
     
     Otherwise (if you do not load all data from tables during erlang app startup) 
-    it is not recommended to set align_lsn to true. In this case to stop PG server stop epgsql replication first.
\ No newline at end of file
+    it is not recommended to set align_lsn to true. In this case to stop PG server stop epgsql replication first.
+    
+## Flow control
+
+It is possible to set `{active, N}` on a TCP or SSL (since OTP 21.3) socket. E.g. for SSL:
```
seriyps, line 145: Let's add a link to https://www.erlang.org/doc/man/inet.html#setopts-2 for those who are not familiar with this option?

doc/streaming.md
```diff
@@ -138,4 +138,39 @@ handle_x_log_data(StartLSN, EndLSN, Data, CbState) ->
...
+        , ssl => true
+        , ssl_opts => [{active, 10}]
+        },
+{ok, Conn} = epgsql:connect(Opts).
+```
+
+It is currently allowed only in the replication mode. Its main purpose is to control the flow of
+replication messages from Postgresql database. If a database is under a high load and a process, which
+handles the message stream, cannot keep up with it then setting this option gives the handling process
+ability to get messages on-demand.
```
seriyps, line 160:
```suggestion
ability to get messages at a pace it is able to handle.
```

doc/streaming.md
```diff
@@ -138,4 +138,39 @@ handle_x_log_data(StartLSN, EndLSN, Data, CbState) ->
...
+
+When the connection is in the asynchronous mode, a process which owns a connection will receive
+```erlang
+{epgsql, Connection, socket_passive}
+```
+The process decides when to activate connection's socket again. To do that it should call:
```
seriyps, line 166:
```suggestion
as soon as underlying socket received N messages from network.

The client process decides when to activate connection's socket again. To do that it should call:
```

src/epgsql_sock.erl
```diff
@@ -313,6 +327,16 @@ format_status(terminate, [_PDict, State]) ->
...
 
 %% -- internal functions --
+-spec send_socket_pasive(pg_sock()) -> pg_sock().
+send_socket_pasive(#state{subproto_state = #repl{receiver = Rec}} = State) when Rec =/= undefined ->
+    Rec ! {epgsql, self(), socket_passive},
+    State;
+send_socket_pasive(#state{subproto_state = #repl{ cbmodule = CbMod
+                                                , cbstate = CbState} = Repl
+                         } = State) ->
+    {ok, NewCbState} = epgsql:handle_socket_passive(CbMod, CbState),
```
seriyps, line 337: I think we should just do `setopts(State, [{active, ValueFromConnectOpts}])` here, because the pace would be synchronously regulated by the execution time of `handle_x_log_data/5` (the same way as it is going to work for any other `epgsql` / `epgsqla` but not `epgsqli` command).

This way we can be sure that the `epgsql`'s mailbox won't overflow if `handle_x_log_data/5` callback takes too long to complete

src/epgsql_sock.erl
```diff
@@ -407,13 +431,29 @@ command_next(#state{current_cmd = PrevCmd,
                         results = []}
     end.
 
-
-setopts(#state{mod = Mod, sock = Sock}, Opts) ->
+setopts(#state{mod = Mod, sock = Sock} = State, DefaultOpts) ->
+    Opts = update_active(State, DefaultOpts),
```
seriyps, line 435: see my note about `#{socket_active => N}` vs `#{tcp_opts := #{active := N}}`

test/epgsql_replication_SUITE.erl
```diff
@@ -1,110 +1,174 @@
...
-     replication_async,
-     replication_sync
-    ].
+  [ connect_in_repl_mode
+  , create_drop_replication_slot
+  , replication_async
+  , replication_sync
+  , replication_async_active_n_socket
+  , replication_sync_active_n_socket
+  ].
```
seriyps, line 36: please let's not re-format everything!

seriyps: I didn't review this file yet because it's not clear what exactly have changed...

seriyps (changes requested):
Main concerns are:

- let's move `active` from `tcp_opts` to the top-level
- let's make `handle_x_log_data/5` control the flow automatically

src/epgsql_sock.erl
```diff
@@ -248,7 +253,11 @@ handle_call({standby_status_update, FlushedLSN, AppliedLSN}, _From,
 handle_call({copy_send_rows, Rows}, _From,
            #state{handler = Handler, subproto_state = CopyState} = State) ->
     Response = handle_copy_send_rows(Rows, Handler, CopyState, State),
-    {reply, Response, State}.
+    {reply, Response, State};
+
+handle_call(activate, _From, State) ->
+    setopts(State, [{active, true}]),
+    {reply, ok, State}.
```
seriyps, line 260: let's make it return what `setopts` returns? Because right now all the `inet:setopts`/`ssl:setopts` call errors are just swallowed

doc/streaming.md
featalion: I think that "messages" would be ambiguous here. It will receive `N` network frames of a size of current `MTU`...

src/epgsql_sock.erl
featalion: If a socket is active, but the callback processes messages slow, it will result in a huge queue on `epgsql` side, right?

test/epgsql_replication_SUITE.erl
featalion: It was 80-90% changed. Don't see a reason to keep the rest unchanged ( ;

src/epgsql_sock.erl
seriyps: By active you mean `{active, true}`? Yes. But `send_socket_passive` wont't be called at all when `{active, true}` because `ssl_passive`/`tcp_passive` won't be ever received.

doc/streaming.md
featalion: Hm, not sure it is a "pace", this is not a rate limiter, but more like "I'm done with this batch, gimme more".

src/epgsql_sock.erl
seriyps: Let's for simplicity assume that one tcp/ssl message contains exactly one X-Log record. And user did `epgsql:connect(#{active => 3, replication => ".."})`.
And let's assume that postgres is able to produce hundreds of X-Log records per second, but `handle_x_log_data/5` can (synchronously) process 1 record per second.

It means epgsql would receive 3 X-Log packets from TCP and then immediately `tcp_passive` msg, so 4 messages in the queue.
It will call `handle_x_log_data/5` 3 times (once for each X-Log packet), it wil ltake 3 seconds.
And then it receives `tcp_passive` and calls `setopts(Sock, [{active, 3}])` and cycle repeats again.

There is never more than 4 messages in the mailbox at the same time.

(note that TCP/SSL messages not necessarily contain exactly one X-Log though)

doc/streaming.md
seriyps: "N network frames of a size of current MTU" is not guaranteed either. By messages I mean rather Erlang messages, not network messages or Postgres protocol messages.

Maybe rephrase as "received N messges from `gen_tcp`/`ssl`, see https://www.erlang.org/doc/man/inet.html#setopts-2 `active` option description"?

doc/streaming.md
```diff
@@ -138,4 +138,39 @@ handle_x_log_data(StartLSN, EndLSN, Data, CbState) ->
...
+## Flow control
+
+It is possible to set `{active, N}` on a TCP or SSL (since OTP 21.3) socket. E.g. for SSL:
+```erlang
+Opts = #{ host => "localhost"
+        , username => "me"
+        , password => "pwd"
+        , database => "test"
+        , ssl => true
+        , ssl_opts => [{active, 10}]
```
seriyps, line 152: "comma-first" style is not in use in `epgsql`. Would be nice to keep it consistent (it's in many places)

test/epgsql_replication_SUITE.erl
```diff
@@ -1,110 +1,174 @@
...
+      case lists:reverse(ReceivedMsgs) of
+        [begin_msg, row_msg | _] -> ok;
+        _ -> error_replication_messages_not_received
+      end;
+    {epgsql, Pid, {x_log_data, _StartLSN, _EndLSN, Msg}} ->
+      [Msg | T] = Pattern,
+      receive_replication_msgs(Module, T, Pid, [row_msg | ReceivedMsgs]);
+    {epgsql, Pid, socket_passive} ->
+      Module:activate(Pid),
+      receive_replication_msgs(Module, Pattern, Pid, ReceivedMsgs)
```
seriyps, line 151: I wonder if this is actually covered. Maybe we should tune the `tcp_opts => [{buffer, X}]` to be small enough to be sure we won't receive everything in just 1 large message?

test/epgsql_replication_SUITE.erl
```diff
@@ -1,110 +1,174 @@
...
 
 replication_sync(Config) ->
-    replication_test_run(Config, ?MODULE).
+  replication_test_run(Config, ?MODULE).
+
+replication_async_active_n_socket(Config) ->
+  replication_test_run(Config, self(), [{tcp_opts, [{active, 1}]}, {ssl_opts, [{active, 1}]}]).
+
+replication_sync_active_n_socket(Config) ->
+  replication_test_run(Config, ?MODULE, [{tcp_opts, [{active, 1}]}, {ssl_opts, [{active, 1}]}]).
```
seriyps, line 66: it looks like this suite only uses plain `gen_tcp` sockets, so, makes no sense to provide `ssl_opts` here? You'd need to add `ssl => true` or better `ssl => required` to make epgsql attempt to do a TLS handshake

test/epgsql_replication_SUITE.erl
```diff
@@ -1,110 +1,174 @@
...
+  epgsql_ct:with_connection(
+    Config,
+    fun(C) ->
+        create_replication_slot(Config, C),
+        %% new connection because main is in the replication mode
+        epgsql_ct:with_connection(
+          Config,
+          fun(C2) ->
+              ExpectedResult = lists:duplicate(length(Queries), {ok, 1}),
+              Res = Module:squery(C2, lists:flatten(Queries)),
```
seriyps, line 83: nit: no need to flatten, epgsql is fine with iolists as SQL

doc/streaming.md
```diff
@@ -138,4 +138,42 @@ handle_x_log_data(StartLSN, EndLSN, Data, CbState) ->
...
+
+It is possible to set `{active, N}` on a [TCP](https://www.erlang.org/doc/man/inet.html#setopts-2)
+or [SSL](https://www.erlang.org/doc/man/ssl.html#setopts-2) (since OTP 21.3) socket. E.g. for SSL:
+```erlang
+Opts = #{host => "localhost",
+         username => "me",
+         password => "pwd",
+         database => "test",
+         ssl => true,
+         ssl_opts => [{active, 10}]
```
seriyps: `socket_active => 10`

src/epgsql_sock.erl
```diff
@@ -189,8 +195,11 @@ set_attr(connect_opts, ConnectOpts, State) ->
 
 %% XXX: be careful!
 -spec set_packet_handler(atom(), pg_sock()) -> pg_sock().
-set_packet_handler(Handler, State) ->
-    State#state{handler = Handler}.
+set_packet_handler(Handler, State0) ->
+    State = State0#state{handler = Handler},
+    Active = get_socket_active(State),
+    setopts(State, [{active, Active}]),
```
seriyps, line 201: Don't understand why do we need it here. Isn't the one in `set_net_socket` enough?

But it won't harm here anyway I guess...

src/epgsql_sock.erl
```diff
@@ -407,13 +435,22 @@ command_next(#state{current_cmd = PrevCmd,
...
         ssl     -> ssl:setopts(Sock, Opts)
     end.
 
+-spec get_socket_active(pg_sock()) -> epgsql:socket_active().
+get_socket_active(#state{handler = H}) when H =/= on_replication ->
+    true;
+get_socket_active(#state{connect_opts = #{socket_active := Active}}) ->
+    Active;
+get_socket_active(#state{connect_opts = Opts}) when is_list(Opts) ->
+    proplists:get_value(socket_active, Opts, true);
```
seriyps, line 450: I know that typespec is confusing, but it is safe to assume `Opts` is always map. It is force-converted to map here:
https://github.com/epgsql/epgsql/blob/39d9662a22018faf0ec0372abc56c55f1420cf63/src/epgsql.erl#L201

And yep, I guess we should eventually have `epgsql:connect_opts_map()`, `epgsql:connect_opts_list()` and `epgsql:connect_opts() :: connect_opts_map() | connect_opts_list()`

src/epgsql_sock.erl
```diff
@@ -407,13 +435,22 @@ command_next(#state{current_cmd = PrevCmd,
...
 setopts(#state{mod = Mod, sock = Sock}, Opts) ->
     case Mod of
         gen_tcp -> inet:setopts(Sock, Opts);
         ssl     -> ssl:setopts(Sock, Opts)
     end.
 
+-spec get_socket_active(pg_sock()) -> epgsql:socket_active().
+get_socket_active(#state{handler = H}) when H =/= on_replication ->
+    true;
+get_socket_active(#state{connect_opts = #{socket_active := Active}}) ->
```
seriyps, line 447: If you do this:

```suggestion
get_socket_active(#state{handler =:= on_replication, connect_opts = #{socket_active := Active}}) ->
```

and assume `connect_opts` is always map, you may drop the 1st and 3rd clauses?

seriyps (commented):
All the new comments are minor. But still need to decide on how do we process `handle_x_log_data/5`

seriyps (approved):
Great, thanks!

Let's see if @davidw have some ideas.

I'll most likely create a follow-up PR some time later to add support for `{active, N}` to all the other commands, not only replication mode (so, `epgsql` and `epgsqla` would behave similar to `handle_x_log_data` while `epgsqli` would start sending `socket_passive`). It would solve the problem with overflows of epgsql mailbox when huge result sets are returned. It won't protect from OOM if someone decides to do `SELECT * FROM one_terabyte_table` (well, it would protect for `epgsqli` interface), but would make the problem more controllable.

src/epgsql.erl
```diff
@@ -571,3 +576,19 @@ to_map(Map) when is_map(Map) ->
     Map;
 to_map(List) when is_list(List) ->
     maps:from_list(List).
+
+%% @doc Activates TCP or SSL socket of the connection.
+%%
+%% If `{active, X}' is set in:
+%% - `tcp_opts' and the current mode is TCP or
+%% - `ssl_opts' and the current mode is SSL
```
seriyps, line 584: seems this doc needs to be updated to reflect the new location and name of this option: `socket_active`

test/epgsql_replication_SUITE.erl
```diff
@@ -1,110 +1,174 @@
...
 handle_x_log_data(StartLSN, EndLSN, Data, CbState) ->
-    {C, Pid} = CbState,
-    Pid ! {epgsql, C, {x_log_data, StartLSN, EndLSN, Data}},
-    {ok, EndLSN, EndLSN, CbState}.
+  {C, Pid} = CbState,
+  Pid ! {epgsql, C, {x_log_data, StartLSN, EndLSN, Data}},
+  {ok, EndLSN, EndLSN, CbState}.
+
+handle_socket_passive({C, _Pid} = CbState) ->
+  spawn(fun() -> epgsql:activate(C) end),
```
seriyps, line 173: this one is not needed now

MikhailKalashnikov (approved):

seriyps:
While trying this option in production, we noticed one thing worth to note: when `epgsql` is in replication mode and stays in passive move for prolonged period of time, PostgreSQL server just closes the connection.

We believe this is directly connected to `wal_sender_timeout` server option https://www.postgresql.org/docs/current/runtime-config-replication.html

So smth like

```
epgsql:squery(C, "SET wal_sender_timeout = '600s'"),
epgsql:start_replication(C, ...)
```

should solve the problem. But we are yet to test it next week. Should probably submit a PR to update the replication docs to mention this nuance.

featalion:
> So smth like
> 
> ```
> epgsql:squery(C, "SET wal_sender_timeout = '600s'"),
> epgsql:start_replication(C, ...)
> ```
> 
> should solve the problem.

I'm not sure it is a problem. A client should process messages in `wal_sender_timeout` seconds, if one sets it to high number, but cannot process data in time it will persist. I would suggest to write a notice to the docs that mentions that the client should handle `sock_closed` from the connection in some way, e.g. reestablish it.

seriyps:
Yup, or we can document both solutions.

PR #278, opened 2022-12-23
Generalize socket active n

### LINKED ISSUE #230: Introduce options in epgsql to deal with potential OOM from statements returning big volumes of data
### Background
As of now, if someone writes a select statement that returns large volumes of data there is a high chance the application node processing this volume might go **OOM**.
This problematic query can in turn kill all the running application nodes if it will be processed by them. For such scenarios the cascading failure of one poorly written select, or a select written when data volumes were too low, can have a catastrophic impact on the whole application cluster. 

One solution to this problem would be to introduce a connection parameter, `max_returned_bytes`, that will be used by epgsql for all the executed queries.
Once the returned bytes exceed the value of the provided option then epgsql will kill the connection process and return back some form of error stating that an `out_of_memory` was reached.

This solution alone won't save us from OOM since there is another scenario:
Let's say epgsql is capable of parsing 10MB/s of messages but our database server plus network is so fast that we received 10GB/s from the socket. In this scenario our mailbox of epgsql gen_sever will get so full that it will cause an OOM.  
After discussing in person with @seriyps, he pointed out the cause of the last OOM was due to the way how we initiate our sockets within epgsql: `{active, true}`.
Had we been using `{active, once}` then we would have been guaranteed to get no more than one TCP message at a time in the mailbox. This option comes with a potential performance hit though.
We can overcome this issue by providing another connection option that allows the callers of epgsql to decide for themselves if they want {active, true} or {active, once}.

### Acceptance Criteria
Introduce two more connection options:
* `max_returned_bytes` -> return an error when the returned handled bytes from epgsql exceeds this number. If no value if provided then continue as today.
* `sock_opts` -> allow the caller to provide sock_opts that epgsql shall use when opening a socket. The default value should be `{active, true}` if no value is provided.

I am open for opinions and further discussion on this topic.

seriyps:
In #275 `socket_active` option have been implemented for the streaming replication mode. In this PR I expand this option usage to any possible epgsql mode.
It may also partially fix #230

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,7 @@ connect(Opts) -> {ok, Connection :: epgsql:connection()} | {error, Reason :: epg
       port =>     inet:port_number(),
       ssl =>      boolean() | required,
       ssl_opts => [ssl:tls_client_option()], % @see OTP ssl documentation
+      socket_active => true | integer(), % @see "Active socket" section below
       tcp_opts => [gen_tcp:option()],    % @see OTP gen_tcp module documentation
       timeout =>  timeout(),             % socket connect timeout, default: 5000 ms
       async =>    pid() | atom(),        % process to receive LISTEN/NOTIFY msgs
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -125,10 +126,12 @@ Only `host` and `username` are mandatory, but most likely you would need `databa
 - `application_name` is an optional string parameter. It is usually set by an application upon
    connection to the server. The name will be displayed in the `pg_stat_activity`
    view and included in CSV log entries.
-- `socket_active` is an optional parameter, which can be true or integer in the range -32768
-  to 32767 (inclusive). This option only works in the replication mode and is used to control
-  the flow of incoming messages. See [Streaming replication protocol](#streaming-replication-protocol)
-  for more details.
+- `socket_active` is an optional parameter, which can be `true` or integer in the range -32768
```
davidw, line 129: 'or an integer'

README.md
```diff
@@ -125,10 +126,12 @@ Only `host` and `username` are mandatory, but most likely you would need `databa
...
+   to 32767 (inclusive, however only positive value makes sense right now).
```
davidw, line 130: 'positive values'

davidw (approved):

seriyps:
Thanks @davidw !

PR #279, opened 2022-12-30
Handle ReadyForQuery synchronously in the replication mode

featalion:
In some cases a message can be received even after a command is complete. One of such examples is when a logical replication is started on a slot already and another client tries to start replication on the same slot. First, PG sends `55006` to indicate the `object in use` error, but right after it the socket receive `Z` (ready for query). At the time of arrival of the `Z` message with value `I`, the command is unset in `epgsql_sock`'s state, which leads to a crash like the following:
```
** Reason for termination ==
** {'module could not be loaded',
       [{undefined,handle_message,
            [90,<<"I">>,
             {state,gen_tcp,#Port<0.99>,<<>>,
                 {214,-1904053072},
                 on_message,
                 {codec,...},
                 {[],[]},
                 undefined,undefined,undefined,undefined,
                 [{<<"application_name">>,<<>>},...],
                 [],[],true,73,undefined,
                 {repl,undefined,0,0,undefined,undefined,undefined,
                     <0.1665.0>,false},
                 #{codecs => [{epgsql_decimal,[]}],...}}]}]}
```
When trying to start a replication, a `ReadyForQuery` message comes after any error.

### DIFF
--- a/src/commands/epgsql_cmd_start_replication.erl
+++ b/src/commands/epgsql_cmd_start_replication.erl
@@ -12,7 +12,6 @@
 
 -type response() :: ok | {error, epgsql:query_error()}.
 
--include("epgsql.hrl").
 -include("protocol.hrl").
 -include("../epgsql_replication.hrl").
 
@@ -65,16 +64,13 @@ execute(Sock, #start_repl{slot = ReplicationSlot, callback = Callback,
 %% CopyBothResponse
 handle_message(?COPY_BOTH_RESPONSE, _Data, Sock, _State) ->
     {finish, ok, ok, epgsql_sock:set_packet_handler(on_replication, Sock)};
-handle_message(?ERROR, #error{codename = object_in_use} = Error, Sock, State) ->
-    %% A replication slot is in use already. In this case, protocol sends a ReadyForQuery message.
+handle_message(?ERROR, Error, Sock, State) ->
+    %% In the case of error, Postgresql replication protocol sends a ReadyForQuery message.
     %% Adds an error to results to handle it later in the ?READY_FOR_QUERY branch.
     Result = {error, Error},
     {add_result, Result, Result, Sock, State};
-handle_message(?ERROR, Error, _Sock, _State) ->
-    Result = {error, Error},
-    {sync_required, Result};
 handle_message(?READY_FOR_QUERY, _Data, Sock, _State) ->
     [Error = {error, _}] = epgsql_sock:get_results(Sock), % assert a single error response
-    {sync_required, Error};
+    {finish, Error, done, Sock};
 handle_message(_, _, _, _) ->
     unknown.
--- a/src/commands/epgsql_cmd_start_replication.erl
+++ b/src/commands/epgsql_cmd_start_replication.erl
@@ -65,8 +65,16 @@ execute(Sock, #start_repl{slot = ReplicationSlot, callback = Callback,
 %% CopyBothResponse
 handle_message(?COPY_BOTH_RESPONSE, _Data, Sock, _State) ->
     {finish, ok, ok, epgsql_sock:set_packet_handler(on_replication, Sock)};
+handle_message(?ERROR, #error{codename = object_in_use} = Error, Sock, State) ->
+    %% A replication slot is in use already. In this case, protocol sends a ReadyForQuery message.
+    %% Adds an error to results to handle it later in the ?READY_FOR_QUERY branch.
+    Result = {error, Error},
+    {add_result, Result, Result, Sock, State};
 handle_message(?ERROR, Error, _Sock, _State) ->
     Result = {error, Error},
     {sync_required, Result};
+handle_message(?READY_FOR_QUERY, _Data, Sock, _State) ->
+    [Error = {error, _}] = epgsql_sock:get_results(Sock), % assert a single error response
+    {sync_required, Error};
 handle_message(_, _, _, _) ->
     unknown.
--- a/test/epgsql_replication_SUITE.erl
+++ b/test/epgsql_replication_SUITE.erl
@@ -10,6 +10,7 @@
 
          connect_in_repl_mode/1,
          create_drop_replication_slot/1,
+         no_replication_slot/1,
          replication_sync/1,
          replication_async/1,
          replication_async_active_n_socket/1,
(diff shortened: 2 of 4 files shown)


seriyps:
Can you add a test that reproduces this issue?

featalion:
I'll try.

featalion:
The test has been added.

test/epgsql_replication_SUITE.erl
```diff
@@ -75,6 +77,37 @@ replication_async_active_n_ssl(Config) ->
...
+        Res1 = Module:start_replication(C, "epgsql_test", Parent, {C, Parent}, "0/0"),
+        ?assertEqual(ok, Res1),
+        spawn(
+          fun() ->
+              %% This connection will be terminated due to an unexpected message:
+              %%   EmptyQueryResponse (B), Byte1('I'), Int32(4)
+              %% https://www.postgresql.org/docs/current/protocol-message-formats.html
+              process_flag(trap_exit, true),
+              C2 = epgsql_ct:connect(Config, User, [{replication, "database"}]),
+              Res2 = Module:start_replication(C2, "epgsql_test", self(), {C2, self()}, "0/0"),
```
seriyps, line 97: I'm not a big expert in replication protocol... Does it mean it's not allowed to have two replications on the same slot?
May you please add some kind of a diagram of what heppens when someone attempts to launch a 2nd replication on the same slot?

The 2nd attempt message exchange would look like this?

```
> SimpleQuery "START_REPLICATION ..."
< CopyBothResponse
< EmptyQueryResponse
```

seriyps: I just can't find `EmptyQueryResponse` anywhere in the docs in context of the replication https://www.postgresql.org/docs/15/protocol-replication.html

featalion: > Does it mean it's not allowed to have two replications on the same slot?

Yes, only one process can occupy a slot.

> The 2nd attempt message exchange would look like this?
> 
> ```
> > SimpleQuery "START_REPLICATION ..."
> < CopyBothResponse
> < EmptyQueryResponse
> ```

Seems like it is. I don't know why it happens, unfortunately. I have tried to find out by googling and monitoring a connection quickly, but I didn't find the reason. Maybe it is a bug in the protocol...

featalion: My mistake, the correct flow is the following:
```
> SimpleQuery "START_REPLICATION ..."
< CopyBothResponse
< ReadyForQuery
```

src/commands/epgsql_cmd_start_replication.erl
```diff
@@ -65,8 +65,16 @@ execute(Sock, #start_repl{slot = ReplicationSlot, callback = Callback,
...
+    %% A replication slot is in use already. In this case, protocol sends a ReadyForQuery message.
+    %% Adds an error to results to handle it later in the ?READY_FOR_QUERY branch.
+    Result = {error, Error},
+    {add_result, Result, Result, Sock, State};
 handle_message(?ERROR, Error, _Sock, _State) ->
     Result = {error, Error},
     {sync_required, Result};
+handle_message(?READY_FOR_QUERY, _Data, Sock, _State) ->
+    [Error = {error, _}] = epgsql_sock:get_results(Sock), % assert a single error response
+    {sync_required, Error};
```
seriyps, line 78: I wonder if it is really in a `sync required` state... What would happen if you return `{finish, Error, done, State}` and then would try to `start_replication/5` to a different, non-occupied slot? Would this 2nd attempt just work or connection is indeed out-of-sync and unusable?

seriyps (commented):
I think the test can be simplified now and also I'm not really convinced that we should set the `sync_required` flag, I think as soon as we got `ready_for_query` we are in a good shape, with `squery` sequence being completed.

test/epgsql_replication_SUITE.erl
```diff
@@ -75,6 +77,38 @@ replication_async_active_n_ssl(Config) ->
...
+  Parent = self(),
+  epgsql_ct:with_connection(
+    Config,
+    fun(C) ->
+        create_replication_slot(Config, C),
+        Res1 = Module:start_replication(C, "epgsql_test", Parent, {C, Parent}, "0/0"),
+        ?assertEqual(ok, Res1),
+        spawn(
+          fun() ->
+              %% This connection will be terminated due to an unexpected message:
```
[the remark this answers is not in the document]
featalion: I thought about it. On one hand it makes perfect sense, on other hand as soon as it is a non-documented behaviour the test case might wait forever (or CT timeout) to complete. If you prefer to have a synchronous test I will update it.

src/commands/epgsql_cmd_start_replication.erl
featalion: It makes sense to me to assume that the connection is in sync here.

seriyps:
On https://www.postgresql.org/docs/current/protocol-replication.html in the description of `START_REPLICATION SLOT slot_name LOGICAL XXX/XXX` it says

> The server can reply with an error, for example if the slot does not exist.

So this behaviour seems to be generic :+1:

src/commands/epgsql_cmd_start_replication.erl
```diff
@@ -65,8 +65,13 @@ execute(Sock, #start_repl{slot = ReplicationSlot, callback = Callback,
...
-handle_message(?ERROR, Error, _Sock, _State) ->
+handle_message(?ERROR, Error, Sock, State) ->
+    %% In the case of error, Postgresql replication protocol sends a ReadyForQuery message.
+    %% Adds an error to results to handle it later in the ?READY_FOR_QUERY branch.
     Result = {error, Error},
-    {sync_required, Result};
+    {add_result, Result, Result, Sock, State};
+handle_message(?READY_FOR_QUERY, _Data, Sock, _State) ->
+    [Error = {error, _}] = epgsql_sock:get_results(Sock), % assert a single error response
+    {finish, Error, Error, Sock};
```
seriyps, line 75: it should be `{finish, Error, done, Sock}`, because we already sent `{error, _}` from line 72. See how `squery` works for the reference

https://github.com/epgsql/epgsql/blob/bcd1aff3be419dee1c94ef447780d91fad9e8f07/src/commands/epgsql_cmd_squery.erl#L76

seriyps (changes requested):
Just onle last bit about `done` (it's important for `epgsqli` interface)

seriyps (approved):
Thanks!

seriyps (approved):

PR #283, opened 2023-04-19
Fix issue with columns in multiple squery response

egobrain:
If squery request contains `select` and than `insert` or `delete` without returning the response of `insert` or `delete` must not contain any columns.

Now:
```sql
select table1_column from table1;
insert into table2 (table2_column) values (1);
delete from table3 where table3_column = 3;
```
will return

```erlang
[
 {ok, [#column{name = <<"table1_column">>}], [{1}]},
 {ok, 1, [#column{name = <<"table1_column">>}]}, []},
 {ok, 1, [#column{name = <<"table1_column">>}]}, []}
]
```

Expected result:
```erlang
[
 {ok, [#column{name = <<"table1_column">>}], [{1}]};
 {ok, 1};
 {ok, 1};
]
```

### DIFF
--- a/src/commands/epgsql_cmd_squery.erl
+++ b/src/commands/epgsql_cmd_squery.erl
@@ -32,7 +32,7 @@
 -record(squery,
         {query :: iodata(),
          columns = [],
-         decoder}).
+         decoder = undefined :: epgsql_wire:row_decoder() | undefined}).
 
 init(Sql) ->
     #squery{query = Sql}.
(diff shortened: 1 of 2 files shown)


egobrain:
> Miht be a good idea to clean the `decoder` field as well, but it's minor

fixed :)

seriyps (approved):

davidw (approved):

PR #284, opened 2023-05-16
Add OTP-26 to CI

seriyps:
See also https://github.com/erlang/otp/issues/7130#issuecomment-1513005012


seriyps:
Ok, it seems there are at least two incompatibilities with OTP-26.
CT logs from the build: https://github.com/epgsql/epgsql/suites/12932403903/artifacts/700838734

1. Unverified server certificates are causing error by default now: `{ssl_negotiation_failed, {options,incompatible, [{verify,verify_peer},{cacerts,undefined}]}}`. I think it can be fixed by setting `{verify, verify_none}` option
2. `inet` port's private protocol we were abusing have changed https://github.com/erlang/otp/issues/7130#issuecomment-1513005012

onno-vos-dev (approved):

Issue #286, opened 2023-07-05
Always check for errors returned from epgsql_sock:send/3


dszoboszlay:
`epgsql_sock` assumes that if a send operation fails and the connection gets closed, it would still receive an `ssl_error` or `tcp_error` message and therefore doesn't always check the return value of `send/3`.

However, erlang/otp#7431 demonstrated that this is not always the case, `ssl:send/2` can fail in a way that it only returns an `{error, Reason}` tuple and the connection gets closed without sending the `ssl_error` message to the controlling process. This leaves `epgsql_sock` in a broken state where it assumes the connection is still up, only idle, while in fact the connection is already closed.

Therefore all calls to `epgsql_sock:send/3` shall check the return value and stop the process.

seriyps:
Yup, sounds reasonable. Just not sure if epgsql should fail gracefully or just crash...

Also, PR would be welcome!

dszoboszlay:
I plan to implement a fix, but opened the issue first to serve as a TODO item for me. :+1:

seriyps:
I think the easiest would be to do

```erlang
case do_send(...) of
  ok -> ok;
  {error, Reason} ->
    throw({stop, {error, Reason}, flush_queue(State#state{sock = undefined}, {error, sock_closed}) }
end
```

here https://github.com/epgsql/epgsql/blob/7ba52768cf0ea7d084df24d4275a88eef4db13c2/src/epgsql_sock.erl#L486 and above

PR #288, opened 2023-08-09
Crash the socket on an ssl:send/2 failure

dszoboszlay:
This is a fix for #286. The gen_tcp case is not modified, because this seems to be specific to ssl (and the way some internal ssl processes relay the messages between the owner of the socket and the underlying gen_tcp socket).

### DIFF
--- a/src/epgsql_sock.erl
+++ b/src/epgsql_sock.erl
@@ -490,7 +490,17 @@ send_multi(#state{mod = Mod, sock = Sock}, List) ->
 do_send(gen_tcp, Sock, Bin) ->
     gen_tcp_send(Sock, Bin);
 do_send(ssl, Sock, Bin) ->
-    ssl:send(Sock, Bin).
+    case ssl:send(Sock, Bin) of
+        ok ->
+            ok;
+        {error, Reason} = Error ->
+            %% In case ssl:send/2 on an active socket fails, the controlling process doesn't get
+            %% notified, see https://github.com/erlang/otp/issues/7431 for reference.
+            %%
+            %% As a workaround, send an ssl_error message on our own.
+            self() ! {ssl_error, Sock, Reason},
+            Error
+    end.
 
 -if(?OTP_RELEASE >= 26).
 gen_tcp_send(Sock, Bin) ->


src/epgsql_sock.erl
```diff
@@ -490,7 +490,17 @@ send_multi(#state{mod = Mod, sock = Sock}, List) ->
...
-    ssl:send(Sock, Bin).
+    case ssl:send(Sock, Bin) of
+        ok ->
+            ok;
+        {error, Reason} = Error ->
+            %% In case ssl:send/2 on an active socket fails, the controlling process doesn't get
+            %% notified, see https://github.com/erlang/otp/issues/7431 for reference.
+            %%
+            %% As a workaround, send an ssl_error message on our own.
+            self() ! {ssl_error, Sock, Reason},
```
seriyps, line 501: 1. Hmm, it will definitely work. But given that `epgsql` tends to grow quite large message queue from time to time, wouldn't it be better to stop stright away (eg, using `throw`)?
2. since now we have started explicitly handling `ssl:send` errors, it might be nice to do that for `gen_tcp` as well. Maybe a separate PR..
3. Would it be possible to create some test that covers this?

seriyps (approved):

src/epgsql_sock.erl
dszoboszlay: 1. I'm not worried that much about the large message queue. I believe that would typically represent data incoming from the DB, correct? So stopping right away would mean losing that unprocessed data, while sending an `ssl_error` message would allow the socket to process it. This is how I would expect `ssl` to work, and it also seems to be the same direction `gen_tcp` is moving, [see the note](https://www.erlang.org/doc/man/gen_tcp.html) on _Non blocking send_ behaviour:

> An background send will detect a 'remote close' and (the inet driver will) mark the socket as 'closed'. [...]
>
> This behaviour is **not** replicated by the socket implementation. A send operation will detect a remote close and immediately return this to the caller, but do nothing else. A reader will therefore be able to extract any data from the OS buffers. If the socket is set to active to not false, the data will be received as expected (`{tcp, ...}` and then a closed message (`{tcp_closed, ...}`) will be received (not an error).

2. I believe with `gen_tcp` a `{tcp_closed, ...}` message will be sent by OTP. I don't have an idea how to get into a similar situation with that protocol. For example tracing the underlying port reveals that even when `ssl:send/2` fails, the `{tcp_closed, ...}` message is sent as expected. The bug is that it won't get translated to an `{ssl_closed, ...}` message by the processes started in the `ssl` layer. So I strongly believe I only need to "patch" `ssl:send/2` to behave similar to `gen_tcp`, and there's nothing to do with the `gen_tcp` case.

3. Yes, I think so. I'll try to come up with something.

dszoboszlay:
@davidw any comments from you? This PR got a bit stalled a month ago.

PR #298, opened 2024-06-27
Update for OTP 27

kivra-pauoli:
## Motivation

Be able to use `epgsql` with Erlang/OTP 27.

At this moment, compilation warnings might prevent consumers from moving forward (if you have "warnings as errors").

## Description

To this end, I:

* prepare the code/CI for OTP 27 compatibility
* update CI as to not run `rebar3_lint` in older versions (23-)
  * I had to update `rebar3_lint` because otherwise it wouldn't work on OTP 27
  * I add a bunch of rule exceptions to `rebar.config` for backward compatibility (solving these exceptions is not a 5-minute task)
* play with `?OTP_RELEASE` so as to not have to bump minimum OTP version
* fix the tests for newer compiler warnings
  * OTP 27 doesn't like `""` as it's used
  * OTP 27 doesn't like `_Var` if the pattern match is being used more than once


.github/workflows/ci.yml
```diff
@@ -27,22 +27,32 @@ jobs:
           - "24.3"
```
seriyps, line 27: Would you mind also bumping OTP 26 to 26.3?

seriyps (approved):
LGTM. Let's see if @davidw has any suggestions. Otherwise will merge & release in a couple of days.

seriyps:
Maybe it's time to drop OTP 21, 22, 23? :)

.github/workflows/ci.yml
kivra-pauoli: Not at all 😄

kivra-pauoli: 9fe1749

seriyps: Ouch, it seems the most recent OTP-26 is 26.2, there were no 26.3 yet, sorry...

kivra-pauoli: Yeah, I could've checked too 😄 No worries, I'll force push to fix the latest commit.

davidw (approved):

kivra-pauoli:
Thanks for merging... would this be published to Hex.pm (and/or GitHub) soonish?

PR #301, opened 2024-09-23
add logging

silverblaze404:
- **added check for integer_overflow**
- **minor fix**
- **Create erlang.yml**
- **Revert "Create erlang.yml"**
- **Return aborted instead of error in with_transaction**
- **rollback fix**
- **Add logging when transaction fails**


seriyps:
What are you trying to achieve?

PR #302, opened 2025-03-26
add monitoring for socket

Alex-erl:


leeyisoft:
how to use "monitoring for socket", please show demo?

Alex-erl:
the problem is that after closing tcp connection messages (tcp_close) do
not come to epgsl_sock, so I had to add a monitor since the monitor fixes
this problem. I have added test into epgsql_replication_SUITE.erl
https://github.com/Alex-erl/epgsql/blob/detached/test/epgsql_replication_SUITE.erl

Test:












































*eplacetion_handle_tcp_close(Config) -> Module = ?config(module, Config),
User = "epgsql_test_replication", SlotName = "epgsql_test", Parent =
self(), ReqRef = make_ref(), ExitRef = make_ref(), Pid = spawn( fun() ->
epgsql_ct:with_connection( Config, fun(C) ->
create_replication_slot(Config, C), Res1 = Module:start_replication(C,
SlotName, Parent, {C, Parent}, "0/0"), ?assertEqual(ok, Res1), SockState =
sys:get_state(C), #{sock := Port} = epgsql_sock:state_to_map(SockState),
Parent ! {ReqRef, Port}, receive ExitRef -> ok after 1000 -> ?assert(false,
"The replication connection thread was not " "killed by a tcp_closed event
within 1000 ms") end end, User, [{replication, "database"}]) end), MonRef =
monitor(process, Pid), %% Receive port from replication handler receive
{ReqRef, Port} -> gen_tcp:close(Port) after 1000 -> ?assert(false, "The
replication handler did not provide an active " "connection port within
1000 ms") end, %% Wait replication connection down receive {'DOWN', MonRef,
process, Pid, sock_closed} -> ok after 1000 -> ?assert(false, "The
replication handler was not " "killed by a tcp_closed event within 1000
ms") end, drop_replication_slot(Config).*


с уважением Александр Свириденко .


Чт, 27 марта 2025 г. в 00:55, leeyi ***@***.***>:

test/epgsql_replication_SUITE.erl
```diff
@@ -17,6 +17,7 @@
          replication_sync_active_n_socket/1,
          replication_async_active_n_ssl/1,
          two_replications_on_same_slot/1,
+         replacetion_handle_tcp_close/1,
```
grahamrhay, line 20: replacetion -> replication?

seriyps:
Hi, thanks for the PR. Are we sure it won't break SSL connection?

seriyps:
I also wonder if getting rid of this hack would solve the problem?

https://github.com/epgsql/epgsql/blob/64aed378455ec3a45c41826896c5561491ff4f7a/src/epgsql_sock.erl#L504-L513

test/epgsql_replication_SUITE.erl
```diff
@@ -131,6 +133,52 @@ no_replication_slot(Config) ->
     "epgsql_test_replication",
     [{replication, "database"}]).
 
+replication_handle_tcp_close(Config) ->
+  Module = ?config(module, Config),
+  User = "epgsql_test_replication",
+  SlotName = "epgsql_test",
+  Parent = self(),
+  ReqRef = make_ref(),
+  ExitRef = make_ref(),
```
seriyps, line 142: I don't get why do you need the "ExitRef". Are you supposed to send it from parent to the `Pid` at some point?

Also would be cool to actually demonstrate how the user can see the `sock_closed` error reason.

Alex-erl: remove ExitRef and fixed ssl

